PR #13, opened 2020-09-03
testing patch for BZ#17909

MichaelChirico:
https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17909

@jeroen is it possible to auto-link bugzilla issues? i'm not familiar with GH bots but happy to look into it if you think it can be done


MichaelChirico:
Merged in https://github.com/wch/r-source/commit/9a165f7ba8ebaeec48937383c49b74bbe8a830da

PR #15, opened 2020-09-15
multi-byte safe string truncations

brodieG:
Continuing process of ensuring that string truncation do not occur 
in the middle of multi-byte characters by using the `R*snprintf_mbcs`
functions.


jeroen:
Looks like this was merged in https://github.com/r-devel/r-svn/commit/d2ad3e6461d003308345ae70f9a3686b8da9a19a 👍

PR #16, opened 2020-09-20
mbcs safe truncations for R*vprintf*

brodieG:
For Rvprintf, only affects Windows when outputing to a
pipe.  REvprintf, only when R_ConsoleFile is not set and
message connection not sunk


jeroen:
Merged by @kalibera in https://github.com/r-devel/r-svn/commit/9d5a6bc97d3792b06fd0d4f07650d15a7b8b67b9

PR #17, opened 2020-09-28
copy srcref for .traceback()

brodieG:
Introduced regression in r76872 by failing to copy
srcref attribute for '.traceback(n)' case.


brodieG:
Merged in with r79305 (478775cb79f5a8)

PR #26, opened 2021-02-01
store path to pkgconfig

jeroen:
This exposes `pkg-config` so it can be used by packages in `Makevars` and `R CMD config` to find external libraries.


s-u:
Do we want to? This is potentially dangerous as the `pkgconfig` may be temporary just for the build and may not exist at run time and/or the toolchain location may vary for distributed binaries. This change woudl probably require all invocations of `pkgconfig` to try twice-with the hard-wired part and with `$PATH`.

jeroen:
It would be really useful on Windows to allow packages to query available system libraries for the correct architecture.

Finding locally installed libraries on Windows is currently very difficult for packages because there are multiple sets of system libraries, for mingw32, mingw64, and soon UCRT. The only reliable way for packages to find the right CPPFLAGS/LDFLAGS for a given system library is checking pkg-config.

In msys2, pkgconfig is part of the toolchain. It is installed by default in `C:/mingw64/bin/` and `C:/mingw32/bin/` alongside gcc and binutils, and ties everything together.

Perhaps on Unix the situation is slightly more complicated, but if we are going to have build servers that build binary R packages both for ARM64 and Intel, you may be in the same boat. 

Yes there are alternative ways like setting PATH or PKG_CONFIG_PATH. But I think because R already has the `R CMD config` system to instruct packages which compilers and flags to use, it makes sense to generalize this so that packages can find external system libraries that are compatible with these same compilers.

jeroen:
> This change woudl probably require all invocations of pkgconfig to try twice-with the hard-wired part and with $PATH.

On most Linux/MacOS configurations, `PKG_CONFIG` will probably just be `pkg-config` so it already gets the one on the PATH, just like with `CC`, `CXX`, etc. I don't think it is needed to do a second attempt.

We would set `PKG_CONFIG` only to a full path on systems that target multiple architectures, like Windows.

PR #38, opened 2021-05-10
leave C0 unchanged

brodieG:


jeroen:
GitHub has an annoying new rule that I need to click "approve" to make the actions run for external contributors. I'll try to add you as a contributor to the project, let's see if that allows us to bypass this.

brodieG:
Great, thanks.

brodieG:
Looks like adding me worked, although it is a bit annoying it is necessary (understandable though, given it does cost something to run all those servers).

PR #44, opened 2021-07-11
consistency fixes to xgettext2pot

MichaelChirico:


MichaelChirico:
cc @jeroen for approval to run the workflow. thanks!

MichaelChirico:
Although, it's weird because this is not in fact my first PR: https://github.com/r-devel/r-svn/pull/13

jeroen:
> Although, it's weird because this is not in fact my first PR: #13

Yes I think GitHub only recognizes you as a contributor when the PR was actually merged, which is not the case of course because this is a mirror.

I have sent you an invite to become a member of this gh organization, so github doesn't require permission anymore.

jeroen:
I'm sorry I messed something up with the member privileges, I sent another invite..

PR #48, opened 2021-07-14
Fix encoding for parsing Sexpr{} output

jeroen:
Sets the correct encoding parsing output from `\Sexpr{}` statements in `Rd` files.


jeroen:
Merged in https://github.com/r-devel/r-svn/commit/ce98e1f78d4c64495a255b9fec4efa69d81c6250

PR #51, opened 2021-08-03
ensure local state in mbrtowc

brodieG:


src/main/character.c
```diff
@@ -612,11 +612,16 @@ substrset(char *buf, const char *const str, cetype_t ienc, int sa, int so,
     } else {
 	/* This cannot work for stateful encodings */
 	if (mbcslocale) {
-	    for (i = 1; i < sa; i++) buf += Mbrtowc(NULL, buf, R_MB_CUR_MAX, NULL);
```
brodieG, line 615: There are two strings being read interleaved.  Each of them requires it's own local state object or else their states will be intermingled which by AFAICT is UB per C99 s7.24.6 p3:

> The initial conversion state corresponds, for a conversion in either
> direction, to the beginning of a new multibyte character in the initial shift
> state. A zero-valued mbstate_t object is (at least) one way to describe an
> initial conversion state. A zero- valued mbstate_t object can be used to
> initiate conversion involving any multibyte character sequence, in any
> LC_CTYPE category setting. If an mbstate_t object has been altered by any of
> the functions described in this subclause, and is then used with a different
> multibyte character sequence, or in the other conversion direction, or with a
> different LC_CTYPE category setting than on earlier function calls, the
> behavior is undefined.299)

Tested with:

    LC_CTYPE=ja_JP.eucjp R
    x <- "abc\xb0\xa1\xb0\xa2\xb0\xa3xyz"
    x
    ## [1] "abc亜唖娃xyz"
    substr(x, 4, 6) <- "\xa6\xa1\xa6\xa2\xa6\xa3"
    x
    ## [1] "abcΑΒΓxyz"

src/main/connections.c
```diff
@@ -4691,6 +4691,8 @@ readFixedString(Rconnection con, int len, int useBytes, Rboolean *warnOnNul)
 
 	p = buf = (char *) R_alloc(R_MB_CUR_MAX*len+1, sizeof(char));
 	memset(buf, 0, R_MB_CUR_MAX*len+1);
+	mbstate_t mb_st;
+	mbs_init(&mb_st);
 	for(i = 0; i < len; i++) {
```
brodieG, line 4696: This is in a utf8 locale, so there should be no state, thus using local state should be safe no matter what.  Also, only used to check validity.  To test (in a UTF-8 locale):

    f <- tempfile()
    writeLines("\xc2\x60", f, useBytes=TRUE)
    stopifnot(inherits(try(readChar(f, 10), silent=TRUE), "try-error"))
    unlink(f)

src/main/printutils.c
```diff
@@ -413,7 +413,7 @@ int Rstrwid(const char *str, int slen, cetype_t ienc, int quote)
 	    unsigned int k; /* not wint_t as it might be signed */
 	    wchar_t wc;
 	    int res = useUTF8 ? (int) utf8toucs(&wc, p):
-		(int) mbrtowc(&wc, p, R_MB_CUR_MAX, NULL);
```
brodieG, line 416: This seems clearly like an oversight, `mb_st` is initialized on line 411, but then unused.  This is also the source of the original bug.

To test:

    LC_CTYPE=en_US.utf8 R
    x <- "\xe9"
    Encoding(x) <- "latin1"
    x

src/main/printutils.c
```diff
@@ -665,7 +665,7 @@ const char *EncodeString(SEXP s, int w, int quote, Rprt_adj justify)
 	for (i = 0; i < cnt; i++) {
 	    wchar_t wc;
 	    int res = (int)(useUTF8 ? utf8toucs(&wc, p):
-			    mbrtowc(&wc, p, R_MB_CUR_MAX, NULL));
```
brodieG, line 668: Same as above.

src/main/printutils.c
```diff
@@ -413,7 +413,7 @@ int Rstrwid(const char *str, int slen, cetype_t ienc, int quote)
 	    unsigned int k; /* not wint_t as it might be signed */
 	    wchar_t wc;
 	    int res = useUTF8 ? (int) utf8toucs(&wc, p):
-		(int) mbrtowc(&wc, p, R_MB_CUR_MAX, NULL);
+		(int) mbrtowc(&wc, p, R_MB_CUR_MAX, &mb_st);
```
brodieG, line 416: This seems clearly like an oversight, `mb_st` is initialized on line 411, but then unused.  This is also the source of the original bug.

To test:

    LC_CTYPE=en_US.utf8 R
    x <- "\xe9"
    Encoding(x) <- "latin1"
    x

src/main/character.c
```diff
@@ -612,11 +612,16 @@ substrset(char *buf, const char *const str, cetype_t ienc, int sa, int so,
...
+	    mbstate_t mb_st_in;
+	    mbs_init(&mb_st_in);
+	    for (i = 1; i < sa; i++)
+		buf += Mbrtowc(NULL, buf, R_MB_CUR_MAX, &mb_st_in);
 	    /* now work out how many bytes to replace by how many */
+	    mbstate_t mb_st_out;
+	    mbs_init(&mb_st_out);
 	    for (i = sa; i <= so && in < strlen(str); i++) {
-		in += (int) Mbrtowc(NULL, str+in, R_MB_CUR_MAX, NULL);
-		out += (int) Mbrtowc(NULL, buf+out, R_MB_CUR_MAX, NULL);
```
brodieG, line 619: There are two strings being read interleaved.  Each of them requires it's own local state object or else their states will be intermingled which by AFAICT is UB per C99 s7.24.6 p3:

> The initial conversion state corresponds, for a conversion in either
> direction, to the beginning of a new multibyte character in the initial shift
> state. A zero-valued mbstate_t object is (at least) one way to describe an
> initial conversion state. A zero- valued mbstate_t object can be used to
> initiate conversion involving any multibyte character sequence, in any
> LC_CTYPE category setting. **If an mbstate_t object has been altered by any of
> the functions described in this subclause, and is then used with a different
> multibyte character sequence, or in the other conversion direction, or with a
> different LC_CTYPE category setting than on earlier function calls, the
> behavior is undefined**.299)

Tested in an non-UTF-8 MBCS locale:

    LC_CTYPE=ja_JP.eucjp R
    x <- "abc\xb0\xa1\xb0\xa2\xb0\xa3xyz"
    x
    ## [1] "abc亜唖娃xyz"
    substr(x, 4, 6) <- "\xa6\xa1\xa6\xa2\xa6\xa3"
    x
    ## [1] "abcΑΒΓxyz"

PR #62, opened 2021-09-08
Retain names in `as.Date.POSIXlt()`

DavisVaughan:
https://bugs.r-project.org/show_bug.cgi?id=18188


DavisVaughan:
Fixed!
https://github.com/wch/r-source/commit/f2a888763cee1e632d515e313ad13d8edc9281c6

PR #69, opened 2021-12-13
Do not assume gcc is on the path

jeroen:


jeroen:
TK says 

```
Please override LOCAL_SOFT if you need that in your builds. That would
be cleaner than assuming that CC is a full path to "gcc" rather than say
"gcc" with an architecture prefix, both of which might make sense under
some circumstances. The Make files are tailored to the currently
recommended toolchain, as has always been the case - and I don't see any
warnings there.

Note that the compiler is added to PATH automatically with Rtools42 and
R installed from the binary installer, and you can influence how with
R_CUSTOM_TOOLS_SOFT and R_CUSTOM_TOOLS_PATH. This is documented e.g. in
the howto. It has to be added manually only when building R from source.
```

PR #78, opened 2022-02-08
Update jitter.Rd

elinw:


eddelbuettel:
@elinw this is a 'read-only' mirror to read. study, test, ... R sources.  

To actually submitt a change you should probably still go via bugzilla.

jeroen:
I think she knows :)

eddelbuettel:
All good then :)

elinw:
Wow a seg fault!

jeroen:
The macos segfault is unrelated to this PR. It will disappear when you rebase on master. (the r checks seems to crash on macos 11.0 so I downgraded the workflow to macos 10.15)

PR #81, opened 2022-03-16
patch from r-devel-vecpat/R/src to r-devel/trunk/src

pmur002:


jeroen:
Merged in https://github.com/r-devel/r-svn/commit/007e31f7cabcb2bda143bfa24b4546c87e01c8ab

PR #83, opened 2022-04-08
remove CTEXT_POP() to fix octal parse data issue

MichaelChirico:
bug: https://bugs.r-project.org/show_bug.cgi?id=18323

### DIFF
--- a/src/main/gram.c
+++ b/src/main/gram.c
@@ -5022,11 +5022,9 @@ static int StringValue(int c, Rboolean forSymbol)
 			octal = 8 * octal + c - '0';
 		    } else {
 			xxungetc(c);
-			CTEXT_POP();
 		    }
 		} else {
 		    xxungetc(c);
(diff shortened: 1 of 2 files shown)


brodieG:
Hey, I haven't looked at the patch in detail (not that my doing so is required in the least), but one thing you'll probably want to do is modify "src/main/gram.y" and then use a [maintainer mode configure](https://cran.r-project.org/doc/manuals/R-admin.html#Maintainer-mode) to generate the gram.c file.

MichaelChirico:
> one thing you'll probably want to do is modify "src/main/gram.y"

:+1: thanks for the pointer & extra thanks for double checking the logic!

brodieG:
[Fixed in trunk with r82185](https://github.com/r-devel/r-svn/commit/cbacc67b6ab9e01f04c50b60f2786e8228bec44a).  Good find/fix @MichaelChirico.

PR #90, opened 2022-07-21
Correct typo in documentation

HayesJohnD:
Changed iff to if.


HayesJohnD:
Please see the proposed attached diff file for changing the original "iff" to "if and only if" for clarity.

[class.tar.gz](https://github.com/r-devel/r-svn/files/9160946/class.tar.gz)

PR #92, opened 2022-07-27
Fix compatibility of katex integration with older browsers

wahlbrink:
Fix compatibility of katex integration with older browsers like IE.


jeroen:
Seems this was fixed in https://github.com/r-devel/r-svn/commit/3cd30f022b0412112eb594260f372ae7095744f3

PR #100, opened 2023-01-04
Add support for S7

t-kalinowski:
This implements the 3 bullet points listed in https://github.com/RConsortium/OOP-WG/issues/222.

> - `@` to avoid the current R7::@ hack.
    One possible approach be to make @ an internal generic when the S4 bit is not set.
    The documentation implies that this is already the case.
> - `inherits()` needs to add branch if what is an `R7_class`.
> -  Ops double dispatch needs to use `Ops.R7_object` if either object inherits from `R7_object` (i.e. the same rule that S4 uses).


src/main/objects.c
```diff
@@ -991,16 +991,29 @@ static SEXP inherits3(SEXP x, SEXP what, SEXP which)
     else
 	PROTECT(klass = R_data_class(x, FALSE));
 
+    int nprot = 1;
+
+    if (inherits(what, "R7_class"))
```
hadley: We might want to make the `isString()` test come first since that will be the common case and should be faster than looking at inheritance.

t-kalinowski: fixed in https://github.com/r-devel/r-svn/pull/100/commits/307253e698199dbc6b334c9fd76534e481f786e7

src/main/attrib.c
```diff
@@ -1613,8 +1613,13 @@ attribute_hidden SEXP do_attrgets(SEXP call, SEXP op, SEXP args, SEXP env)
 	    PROTECT(obj = shallow_duplicate(obj));
 	else
 	    PROTECT(obj);
+
+	if (IS_S4_OBJECT(obj)) {
 	check_slot_assign(obj, input, value, env);
 	obj = R_do_slot_assign(obj, input, value);
```
DavisVaughan, line 1617: Looks like we are missing tabs here

And `setAttrib` below looks to be tabbed one too many times

src/main/attrib.c
```diff
@@ -1613,8 +1613,13 @@ attribute_hidden SEXP do_attrgets(SEXP call, SEXP op, SEXP args, SEXP env)
...
+	} else {
+		setAttrib(obj, input, value);
```
DavisVaughan: This isn't strictly necessary for R7 right? The `DispatchOrEval()` bit above means that our `@<-` R7 method already works

This really goes against the spirit of the S4 replacement operator, as documented:

> The replacement operator checks that the slot already exists on the object (which it should if the object is really from the class it claims to be).

i.e. this change means that we don't do that check for non-S4 objects.

I think this will be easier to get in if we change as little as possible.

src/main/eval.c
```diff
@@ -3847,6 +3847,11 @@ int DispatchGroup(const char* group, SEXP call, SEXP op, SEXP args, SEXP rho,
 		     (streql(rname, "+.POSIXt") || streql(rname, "+.Date")) )
 		lsxp = R_NilValue;
 
+		else if(inherits(CAR(args), "R7_object"))
+			rsxp = R_NilValue;
+		else if (inherits(CADR(args), "R7_object"))
+			lsxp = R_NilValue;
```
DavisVaughan: Need to match tabbing and `if (` spacing with what is already there

DavisVaughan: Might want a comment like what they have for difftime

```
/* special-case some methods involving difftime */
```

would be useful to describe how you can get here, i.e. something like

"when both an R7 and a non-R7 object define a method, prefer the R7 method"

It seems worth mentioning because the case of 2 R7 objects means that both `lsxp` and `rsxp` will be identical and you won't get here.

src/main/objects.c
```diff
@@ -991,16 +991,32 @@ static SEXP inherits3(SEXP x, SEXP what, SEXP which)
     else
 	PROTECT(klass = R_data_class(x, FALSE));
 
-    if(!isString(what))
-	error(_("'what' must be a character vector"));
+    int nprot = 1;
```
DavisVaughan: Probably slightly more natural to put `int nprot = 0` at the top of the function and then increment it by 1 after the above branch

src/main/objects.c
```diff
@@ -991,16 +991,32 @@ static SEXP inherits3(SEXP x, SEXP what, SEXP which)
...
+
+	if (!isString(what))
```
DavisVaughan: Indentation looks off here

src/main/objects.c
```diff
@@ -991,16 +991,32 @@ static SEXP inherits3(SEXP x, SEXP what, SEXP which)
...
+	{
+		if (inherits(what, "R7_class"))
+		{
+			SEXP R7_ns = R_FindNamespace(mkString("R7"));
+			SEXP cl = PROTECT(lang2(install("R7_class_name"), what));
+			what = PROTECT(eval(cl, R7_ns));
```
DavisVaughan: Do we want to rely on an R7 R callback like this here?

Would it be an easier sell if this was "R7 agnostic" C code that just grabbed the `name` and `package` attributes off of the R7_class object and smushed them together separated by `::`?

I feel like R core is going to be hesitant to allow R's C code to call out to R code managed by a package on CRAN that in theory could be updated at any time to something incompatible with this

DavisVaughan: If we do keep this, we could consider making `R7_ns` and `cl` static objects so we don't have to look them up every time (only once)

src/main/objects.c
```diff
@@ -991,16 +991,32 @@ static SEXP inherits3(SEXP x, SEXP what, SEXP which)
...
     int j, nwhat = LENGTH(what);
 
-    if( !isLogical(which) || (LENGTH(which) != 1) )
+	if (!isLogical(which) || (LENGTH(which) != 1))
 	error(_("'which' must be a length 1 logical vector"));
     Rboolean isvec = asLogical(which);
 
-    if(isvec)
+    if (isvec)
+	{
```
DavisVaughan: some weird fiddly bits with the indentation here - i feel like we need to get your RStudio preferences synced with what R core uses? This works for me

Screen Shot 2023-02-12 at 5 11 08 PM

src/main/subset.c
```diff
@@ -647,7 +647,7 @@ static int ExtractExactArg(SEXP args)
 }
 
 /* Version of DispatchOrEval for "[" and friends that speeds up simple cases.
-   Also defined in subassign.c */
+   Also defined in subassign.c and attrib.c* */
```
DavisVaughan:
```suggestion
   Also defined in subassign.c and attrib.c */
```

src/main/subassign.c
```diff
@@ -1520,7 +1520,7 @@ static R_INLINE int SubAssignArgs(SEXP args, SEXP *x, SEXP *s, SEXP *y)
 }
 
 /* Version of DispatchOrEval for "[" and friends that speeds up simple cases.
-   Also defined in subset.c */
+   Also defined in subset.c and attrib.c*/
```
DavisVaughan:
```suggestion
   Also defined in subset.c and attrib.c */
```

src/main/attrib.c
```diff
@@ -1860,36 +1865,69 @@ SEXP R_do_slot_assign(SEXP obj, SEXP name, SEXP value) {
...
+        name = installTrChar(STRING_ELT(name, 0));
+
+    if (IS_S4_OBJECT(object))
+    {
+        if (!s_dot_Data)
+            init_slot_handling();
+        ans = R_do_slot(object, name);
+        UNPROTECT(1);
+        return ans;
     }
```
DavisVaughan, line 1924: Yea I think I would try to avoid changing all this and focus on just adding `DispatchOrEval()` support for `@` "to align it with existing documentation"

DavisVaughan (commented):
I think that when we submit this to base R, it should be in the form of 4 small PRs that will be easier for them to digest:

- Adding S3 generic support to `%*%` (S7 agnostic)
- Making `@` support `DispatchOrEval()` as it is documented (S7 agnostic)
- Making `inherits()` aware of S7 objects
- Making `Ops` aware of S7 objects

Submitting all 4 at once in the same patch would probably slow things down a lot.

I would also probably try to avoid changing existing code as much as possible. Like the changes to `do_AT()` could probably be simplified greatly to just focus on `DispatchOrEval()`. Ignoring everything else for now seems like the best way to actually get the changes into R.

Some other things I noticed that should be updated:

- [ ] Move to S7 not R7
- [ ] Update styling to match base R (match `if(` spacing, match 8 space tabs). I think we might need to get your RStudio IDE (or whatever IDE you use) synced with what base R uses 😄 , I have a comment below about the specific settings i used that worked for me.
- [ ] Update `@<-` docs to not be so S4 specific. Probably need to update `@` docs to mention that it works on non-S4 objects.
- [ ] Update `InternalGenerics` docs to know about `@` being an internal generic https://github.com/wch/r-source/blob/37e2fd10b9ec771040e755691e4d10f7068c3431/src/library/base/man/InternalMethods.Rd#L56
- [ ] Update docs of `%*%` to mention that it now supports S3 methods (The `Details` section says "This operator is S3 generic but not S3 generic...")
- [ ] Look into updating `getGroupMembers()` to be aware of `%*%`? That might not be possibly since it isn't really a true "group op"?

src/main/objects.c
mmaechler: Yes,  the length-8 Tab  has been really standard in R's C code.

src/library/base/R/zzz.R
```diff
@@ -226,6 +225,8 @@ assign("as.integer", function(x, ...) UseMethod("as.integer"),
        envir = .GenericArgsEnv)
 assign("as.logical", function(x, ...) UseMethod("as.logical"),
        envir = .GenericArgsEnv)
+assign("%*%", function(x, y) UseMethod("%*%"),
+       envir = .GenericArgsEnv)
```
mmaechler, line 231: IIUC we won't do this,  but rather keep %*% primitive  and use DispatchOrEval() in its C code.
I'm pretty sure the above change would be disruptive,  also a performance killer in some important cases.

src/main/objects.c
t-kalinowski: Thank you, this tip helped a lot!

This is my current understanding of the conventions for indentation in R's C code:

Blocks of code that belong to things `if`, `else`, `for`, function bodies etc, have one additional indentation level relative to their parent expression, where an "indentation level" corresponds to 4 spaces (occasionally 2 spaces). 

After indentation level of a line is established, strings of 8 spaces are replaced with a single tab character, starting from the left.

E.g., if I set my tab-display-width to 8 spaces and make whitespace characters visible, this is what I see in current master branch `do_AT()`: 
image

Taking most of this body and placing it under an `if(IS_S4_OBJECT(object))` (as was recommended in the meeting earlier today), I arrive at this:

image

eddelbuettel: There is a short (possible old) description near the end of the [R Coding Standards](https://rstudio.github.io/r-manuals/r-ints/R-coding-standards.html) section of the R Internals manual.  This also has the ESS config snippet some of us have used for decades :)

src/main/eval.c
t-kalinowski: Seems like the current convention is to alternative `if(` and `if (` every other expression :laughing: 
image

t-kalinowski: When in Rome...

src/library/base/R/zzz.R
t-kalinowski: Thank you @mmaechler for the review!

I reverted this line. Unfortunately, it causes one test to fail. Specifically, the test that checks that none of the symbols in `.ArgsEnv` are generic: https://github.com/r-devel/r-svn/blob/c5f16db96ad76504893a1070d3b46b4093bb49c4/tests/primitives.R#L30-L48

Digging a little, my understanding is that `.GenericArgsEnv` and `.ArgsEnv` are only used for (infrequent) code introspection and in tests, and that the `UseMethod()` call in the `body()` of closures in `.GenericArgsEnv` is not actually evaluated during normal code execution.

Note that the other primitive generics that dispatch to Ops (and similarly are performance sensitive) are also assigned to `.GenericArgsEnv` in the same manner (as a closure with a `UseMethod()` body): 
https://github.com/r-devel/r-svn/blob/e22e2a7c353d27134fcfcbbaee9c9c64952f772e/src/library/base/R/zzz.R#L192-L193

I've reverted the reversion for now, and am open for other suggestions for about how to proceed. (e.g., special case an exception for `%*%` in the linked test instead of binding a dummy closure to `%*%` in `.GenericArgsEnv`).

src/main/objects.c
t-kalinowski: This was discussed in the working group meeting--consensus was reached that accessing the S7 namespace is the preferred approach.
I updated the implementation by making the SEXPs static. Thank you for the suggestion!

src/library/base/R/zzz.R
t-kalinowski: Thank you for the review @DavisVaughan! 

Responses in line:

> I think that when we submit this to base R, it should be in the form of 4 small PRs that will be easier for them to digest:
> 
>   * Adding S3 generic support to `%*%` (S7 agnostic)
> 
>   * Making `@` support `DispatchOrEval()` as it is documented (S7 agnostic)
> 
>   * Making `inherits()` aware of S7 objects
> 
>   * Making `Ops` aware of S7 objects
> 
> 
> Submitting all 4 at once in the same patch would probably slow things down a lot.

Good idea! I will submit this as 4 separate patches.


> I would also probably try to avoid changing existing code as much as possible. Like the changes to `do_AT()` could probably be simplified greatly to just focus on `DispatchOrEval()`. Ignoring everything else for now seems like the best way to actually get the changes into R.

I've updated `do_AT()` to minimize the size of the diff by collecting all the new behavior into an `if` branch that returns early from the function, and leaving as much as possible of the existing code as-is. I've also removed the default fallback of `getAttrib()` in `@` and `setAttrib()` in `@<-`, leaving an error message instead. 
 
> Some other things I noticed that should be updated:
> 
>   * [x]  Move to S7 not R7

Completed via https://github.com/RConsortium/OOP-WG/pull/276


>   * [x]  Update styling to match base R (match `if(` spacing, match 8 space tabs). I think we might need to get your RStudio IDE (or whatever IDE you use) synced with what base R uses smile , I have a comment below about the specific settings i used that worked for me.

Regarding indentation whitespace, I've (manually) applied the following rule to the patch:
- each additional indentation level is 4 spaces
- strings of 8 leading spaces are replaced by a tab character.

Regarding `if (` vs `if(`, I'm not sure there is much of any convention. 
```bash
$ grep -roh 'if (' src/main | wc -l
9723
$ grep -roh 'if(' src/main | wc -l
4009
```
The current state of the code reflects my best effort to adapt to the local style, but even that is difficult to reliably infer what is preferred.

For posterity, this was also discussed in:
- https://github.com/r-devel/r-svn/pull/100#discussion_r1103884478
- https://github.com/r-devel/r-svn/pull/100#discussion_r1103881078

If there is still appetite for discussions about formatting, I would direct efforts towards
https://github.com/r-devel/r-svn/pull/101


>   * [x]  Update `@<-` docs to not be so S4 specific. Probably need to update `@` docs to mention that it works on non-S4 objects.

Done.

>   * [x]  Update `InternalGenerics` docs to know about `@` being an internal generic https://github.com/wch/r-source/blob/37e2fd10b9ec771040e755691e4d10f7068c3431/src/library/base/man/InternalMethods.Rd#L56

Done.
 
>   * [x]  Update docs of `%*%` to mention that it now supports S3 methods (The `Details` section says "This operator is S3 generic but not S3 generic...")

Done.
 
>   * [ ]  Look into updating `getGroupMembers()` to be aware of `%*%`? That might not be possibly since it isn't really a true "group op"?

For posterity, this was discussed in the most recent OOP working group meeting. Three alternatives discussed were:
 - Defining a new `Matrix` group, (maybe with `t()` `crossprod()` and `tcrossprod()`).
 - Adding `%*%` to the `Arith` group.
 - Doing nothing.

We decided that the group definitions in the methods package will not be updated as part of this patch.

t-kalinowski:
Thank you for the review @DavisVaughan! 

Responses in line:

> I think that when we submit this to base R, it should be in the form of 4 small PRs that will be easier for them to digest:
> 
>   * Adding S3 generic support to `%*%` (S7 agnostic)
> 
>   * Making `@` support `DispatchOrEval()` as it is documented (S7 agnostic)
> 
>   * Making `inherits()` aware of S7 objects
> 
>   * Making `Ops` aware of S7 objects
> 
> 
> Submitting all 4 at once in the same patch would probably slow things down a lot.

Good idea! I will submit this as 4 separate patches.


> I would also probably try to avoid changing existing code as much as possible. Like the changes to `do_AT()` could probably be simplified greatly to just focus on `DispatchOrEval()`. Ignoring everything else for now seems like the best way to actually get the changes into R.

I've updated `do_AT()` to minimize the size of the diff by collecting all the new behavior into an `if` branch that returns early from the function, and leaving as much as possible of the existing code as-is. I've also removed the default fallback of `getAttrib()` in `@` and `setAttrib()` in `@<-`, leaving an error message instead. 
 
> Some other things I noticed that should be updated:
> 
>   * [x]  Move to S7 not R7

Completed via https://github.com/RConsortium/OOP-WG/pull/276


>   * [x]  Update styling to match base R (match `if(` spacing, match 8 space tabs). I think we might need to get your RStudio IDE (or whatever IDE you use) synced with what base R uses smile , I have a comment below about the specific settings i used that worked for me.

Regarding indentation whitespace, I've (manually) applied the following rule to the patch:
- each additional indentation level is 4 spaces
- strings of 8 leading spaces are replaced by a tab character.

Regarding `if (` vs `if(`, I'm not sure there is much of any convention. 
```bash
$ grep -roh 'if (' src/main | wc -l
9723
$ grep -roh 'if(' src/main | wc -l
4009
```
The current state of the code reflects my best effort to adapt to the local style, but even that is difficult to reliably infer what is preferred.

For posterity, this was also discussed in:
- https://github.com/r-devel/r-svn/pull/100#discussion_r1103884478
- https://github.com/r-devel/r-svn/pull/100#discussion_r1103881078

If there is still appetite for discussions about formatting, I would direct efforts towards
https://github.com/r-devel/r-svn/pull/101


>   * [x]  Update `@<-` docs to not be so S4 specific. Probably need to update `@` docs to mention that it works on non-S4 objects.

Done.

>   * [x]  Update `InternalGenerics` docs to know about `@` being an internal generic https://github.com/wch/r-source/blob/37e2fd10b9ec771040e755691e4d10f7068c3431/src/library/base/man/InternalMethods.Rd#L56

Done.
 
>   * [x]  Update docs of `%*%` to mention that it now supports S3 methods (The `Details` section says "This operator is S3 generic but not S3 generic...")

Done.
 
>   * [ ]  Look into updating `getGroupMembers()` to be aware of `%*%`? That might not be possibly since it isn't really a true "group op"?

For posterity, this was discussed in the most recent OOP working group meeting. Three alternatives discussed were:
 - Defining a new `Matrix` group, (maybe with `t()` `crossprod()` and `tcrossprod()`).
 - Adding `%*%` to the `Arith` group.
 - Doing nothing.

We decided that the group definitions in the methods package will not be updated as part of this patch.

PR #101, opened 2023-01-04
Clang format

t-kalinowski:


t-kalinowski:
One downside of auto-formatting is that you lose a useful `git blame`.
One "solution" is to rewrite history, applying formatting to each individual commit.
As a curiosity, I tried that here: https://github.com/t-kalinowski/r-svn-formatted

PR #103, opened 2023-01-25
Change in graphics::persp documentation example codes owing to a bug report

itsdebartha:
This is a change in documentation, owing to the bug report at https://bugs.r-project.org/show_bug.cgi?id=17699. Changes made:
- modify the function f so that it does not return NAs
- remove the scale of the function f
- modify the expand parameter of persp to better represent the changes in the graph
- modify the z-axis values of trans3d in both points and lines

### DIFF
--- a/src/library/graphics/man/persp.Rd
+++ b/src/library/graphics/man/persp.Rd
@@ -151,13 +151,14 @@ require(grDevices) # for trans3d
 
 x <- seq(-10, 10, length.out = 30)
 y <- x
-f <- function(x, y) { r <- sqrt(x^2+y^2); return(ifelse(is.na(sin(r)/r), 1, sin(r)/r)) }
+f <- function(x, y) { r <- sqrt(x^2+y^2); sin(r)/r }
 z <- outer(x, y, f)
 op <- par(bg = "white")
-persp(x, y, z, theta = 30, phi = 30, expand = 0.75, col = "lightblue")
-persp(x, y, z, theta = 30, phi = 30, expand = 0.75, col = "lightblue",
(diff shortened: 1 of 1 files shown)


hturner:
Thinking about it some more, if we know x and y are not near zero - as we do when the sequence is even length - we can keep it simple and skip handling NAs. So I think you can just have `f <- function(x, y) { r <- sqrt(x^2+y^2); sin(r)/r }`.

Modifying the `expand` argument doesn't solve the issue of the z axis labels getting too close to the z axis title when the scaling is removed. I suggest instead adding `cex.axis = 0.7, cex.lab = 0.7` to make all the axis text a little smaller in the second `persp` plot.

itsdebartha:
@hturner in that case, we have to check if the length of `x` is even or not and return `f` accordingly. Will it be better than this one as that will involve one more line of checking (i.e. checking if the length of `x` is even or not)?

hturner:
I think as the function is just created for the purposes of this example, and we define x and y in the example, then we can ignore the fact that it is undefined for x = y = 0. That is, I don't think the function f() needs to handle this case, as it doesn't arise in the example.

itsdebartha:
That sounds great...
I'll do the changes

hturner:
Great. You can always make a comment when you submit to Bugzilla that handling NAs is unnecessary for this example, but if NAs are to be handled it would be neater to do this inside f() since that follows the definition of the sinc function. Then the R Core member that reviews the patch can decide which solution they prefer.

itsdebartha:
The commit 69abfba1aa7db234d0e34eaae1e4bfc57f7bc081 addresses this issue and solves the problem

PR #110, opened 2023-02-28
fix errors: `dput(<S7_object>)` ; `debug(<S7_method>)`

t-kalinowski:
https://github.com/RConsortium/OOP-WG/issues/272
https://github.com/RConsortium/OOP-WG/issues/281


jeroen:
Maybe add a test?

t-kalinowski:
Unfortunately, adding an R test isn't straightforward. 

S7 objects are a unique object type where the SEXP type is `S4SXP`, but the S4 object bit is not set. I.e., where `!IS_S4_OBJECT(x) && TYPEOF(x) == S4SXP`. This is not a valid type of R object that can be created from plain R code. (Prior to R 2.4, there used to be [valid S4 objects](https://github.com/wch/r-source/blob/e7a21904029917a63b4717b53a173b01eeabcc7b/src/main/objects.c#L1689-L1698) like this, but [not anymore](https://github.com/wch/r-source/blob/e7a21904029917a63b4717b53a173b01eeabcc7b/src/main/saveload.c#L2122-L2124)).  

The S7 package creates this special SEXP object in C, but we're not at the stage of taking a dependency on S7 in R itself yet. Unfortunately, without a pathway to creating these special SEXP objects, there is no way I see to add tests that can live in r-svn trunk.

t-kalinowski:
Patch submitted to bugzilla: https://bugs.r-project.org/show_bug.cgi?id=18511

PR #111, opened 2023-02-28
[BZ-18480]: Enhancements to `dendrapply`

ahl27:
Various enhancements to `dendrapply`, as described in the original repository: https://github.com/ahl27/new_dendrapply
Implementation is tested and working against all unit tests in `dendextend`, should be a drop-in replacement for current version of `dendrapply`. R code is implemented as shown to preserve all the quirks of the current version so as not to break older packages.

Improvements:
- 2-3x speedup over original `stats::dendrapply`
- No recursion, so no stack overflow errors with large/deep trees
- New optional argument `how=c('pre.order', 'post.order')` to apply the function in pre-order or post-order traversal
  - default value 'pre.order' applies identically to original `stats::dendrapply` (node evaluated before its children)
  - new option 'post.order' applies to all children of node before node itself
 

edit:  bugzilla report opened at https://bugs.r-project.org/show_bug.cgi?id=18480


ahl27:
Edit: I've re-enabled this test. See next comment for details; keeping this comment to track history etc.

Original comment:

I've disabled a test that's failing:
```
## labels.dendrogram(), dendrapply(), etc -- see comment #15 of PR#15215 :
(D <- as.dendrogram(hclust(dist(cbind(setNames(c(0,1,4), LETTERS[1:3]))))))
stopifnot(
    identical(labels(D), c("C", "A", "B")),
     ## has been used in "CRAN package space"
    identical(suppressWarnings(dendrapply(D, labels)),
              list("C", list("A", "B"), "C")))
```

I'm not sure how to replicate this functionality. Currently thinking about the best approach for this. I'm honestly not entirely sure why this even performs like it does in the first place...going to work through it slowly later, but suggestions/ideas are definitely welcome. 

Expected Output:
```
[[1]]
[1] "C"

[[2]]
[[2]][[1]]
[1] "A"

[[2]][[2]]
[1] "B"

[[3]]
[1] "C"
```

Actual Output (new function):
```
[[1]]
[1] "C"

[[2]]
[[2]][[1]]
[1] "A"

[[2]][[2]]
[1] "B"
```

I'm pretty sure it stems from the `r[] <- lapply(d, Napply)` call in `stats::dendrapply`, where the results of `lapply(d, Napply)` may be replicated partially if the length of `r` is longer than that of the `lapply` result. This is a similar scenario to something like:
```
v <- numeric(9)
v[] <- c(1,2)
```

Edit with explanation: The reason for this result is because of how the original `stats:dendrapply` is written. The function first applies `labels` to the root and converts it to a list, giving this output:
```
[[1]] "C"
[[2]] "A"
[[3]] "B"
```
We then apply the function to the root nodes, producing this output:
```
[[1]] "C"
[[2]][[1]] "A"
[[2]][[2]] "B"
```

However, the original list generated from the node is length 3. Thus, as R assigns the result to the original list, R will implicitly replicate the result across the length until it runs out of space, just like if you do the example using `v` from before. This results in the value `"C"` being replicated to the third position, producing the observed output.

ahl27:
I've re-enabled the failing test. The C code will now accurately reproduce behavior from `stats::dendrapply` when doing things like: 
```
D <- as.dendrogram(hclust(dist(cbind(setNames(c(0,1,4), LETTERS[1:3])))))
dendrapply(D, labels)
```

Behavior should be identical, with the small exception that it doesn't throw a warning where the previous `stats::dendrapply` did. Warning the user is a little trickier because of the way the code runs, but it could be doable if required.

ahl27:
Merged upstream and restarted checks since miktex download seems to be working again

ahl27:
Fixed some edge cases thanks to comments from Ivan Kyrlov. R logic now correctly accounts for cases where the function returns a result that is not list-like and/or is shorter than the node without unclassing the dendrogram object.

ahl27:
Lots of updates following work at the R project sprint 2023. Runtime is improved, now 3-5x faster than base R. Should be a drop-in replacement for the current dendrapply.

PR #114, opened 2023-03-14
Fix three issues with ALTREP lists

gaborcsardi:
- The `VECTOR_ELT_0` and `SET_VECTOR_ELT_0` macros need to use `STDVEC_DATAPTR` to avoid the ALTREP dispatch.
- `SET_VECTOR_ELT()` needs to use `VECTOR_ELT()` and not `VECTOR_ELT_0()` when adjusting the refcount of the element that was removed.
- Make VECSXP also wrappable in `wrap_meta()`.


jeroen:
You can fix CI by rebasing on master.

gaborcsardi:
This has been merged.

PR #116, opened 2023-03-14
Allow Dataptr_or_null method in ALTLIST classes

gaborcsardi:
This can be very useful to iterate over an ALTLIST
very quickly. For example, in an ALTLIST class
that implements a sublist, no copying is needed.

This method always returns a const pointer, so
it cannot be used to change the list.

### DIFF
--- a/src/main/altrep.c
+++ b/src/main/altrep.c
@@ -351,6 +351,11 @@ R_xlen_t /*attribute_hidden*/ ALTREP_TRUELENGTH(SEXP x) { return 0; }
 
 static R_INLINE void *ALTVEC_DATAPTR_EX(SEXP x, Rboolean writeable)
 {
+    /* Disallow taking the `DATAPTR()` of an ALTLIST. This check could be
+       moved to `DATAPTR()` to catch more faulty usages. */
+    if (TYPEOF(x) == VECSXP && writeable)
+        ALTREP_ERROR_IN_CLASS("cannot take a writeable DATAPTR of an ALTLIST", x);
+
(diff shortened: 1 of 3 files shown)


src/main/altrep.c
```diff
@@ -351,6 +351,10 @@ R_xlen_t /*attribute_hidden*/ ALTREP_TRUELENGTH(SEXP x) { return 0; }
 
 static R_INLINE void *ALTVEC_DATAPTR_EX(SEXP x, Rboolean writeable)
 {
+    /* protect VECSXP, it can only be changed via SET_VECTOR_ELT() */
```
lionel-:
```suggestion
    /* Disallow taking the `DATAPTR()` of an ALTLIST. This check could be
       moved to `DATAPTR()` to catch more faulty usages. */
```

src/main/altrep.c
```diff
@@ -351,6 +351,10 @@ R_xlen_t /*attribute_hidden*/ ALTREP_TRUELENGTH(SEXP x) { return 0; }
...
+    if (TYPEOF(x) == VECSXP && writeable)
+        ALTREP_ERROR_IN_CLASS("ALTREP lists must not return a writeable DATAPTR", x);
```
lionel-:
```suggestion
        ALTREP_ERROR_IN_CLASS("cannot take a writeable DATAPTR of an ALTLIST", x);
```

lionel- (commented):
It's probably better to merge the two commits in order to produce a single patch.

gaborcsardi:
This has been merged.

PR #125, opened 2023-05-02
throttle calls to R_CheckUserInterrupt

jeroen:
See: https://stat.ethz.ch/pipermail/r-devel/2023-May/082597.html


s-u:
Interesting idea, two issues that may be worth considering:

- interactive sessions that run the system event loop may need a lower threshold since 10Hz is too slow and will affect user experience
- `currentTime()` is not entirely free - it adds about 0.1µs per call on Linux (and about 1/3 of that on macOS) so it would be good to know if there are tight-spinning loops where that may matter.

PR #128, opened 2023-06-07
define OBJSXP

t-kalinowski:
- add `OBJSXP` as an alias for `S4SXP` 
- make `dput()` and `print()` do something sensible on an `OBJSXP` when the S4 bit is not set
- `typeof()` returns `"object"` on an `OBJSXP` if the S4 bit is unset
- `.Internal(inspect(x))` prints `OBJSXP` instead of `S4SXP` if the S4 bit is unset

~~Also, this PR adds a `base::object()` constructor that returns an `OBJSXP` with the S4 bit unset. This is mainly for testing and experimentation purposes, and I'm not currently thinking it should be added to base R.~~ The code for the constructor is commented out, since it's not clear if this is a good thing to add to base R presently.


t-kalinowski:
This is merged into master now.

PR #133, opened 2023-08-11
proposed patch for user-friendliness of methods dispatch failure

MichaelChirico:
ref: https://stat.ethz.ch/pipermail/r-devel/2023-August/082789.html


MichaelChirico:
Merged: 241717103f2a382285d53ae5d745fe647a06fa51

PR #134, opened 2023-08-28
fix matrixOps signature in tools pkg checks

t-kalinowski:
Updates the signature of the matrixOps generic in the tools static analysis routines to the actual signature of matrixOps as it is documented, from `e1, e2` to `x, y`.

Otherwise, defining a `matrixOps.foo <- function(x, y) {....}` method causes an R CMD check warning like this:
```
W  checking S3 generic/method consistency ...
   matrixOps:
     function(e1, e2)
   matrixOps.S7_object:
     function(x, y)
   See section ‘Generic functions and methods’ in the ‘Writing R
   Extensions’ manual.
```

As seen in https://github.com/RConsortium/OOP-WG/pull/328

### DIFF
--- a/src/library/tools/R/utils.R
+++ b/src/library/tools/R/utils.R
@@ -1746,7 +1746,7 @@ function(parent = parent.frame())
            envir = env)
     assign("Ops", function(e1, e2) UseMethod("Ops"),
            envir = env)
-    assign("matrixOps", function(e1, e2) UseMethod("matrixOps"),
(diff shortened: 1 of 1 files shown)


mmaechler (approved):

mmaechler:
Fixed in R's sources (svn rev 85050)

jeroen:
Fixed in https://github.com/r-devel/r-svn/commit/b0abc0edbe4ae1a61ee95bb1f049014b8738a2dd

PR #137, opened 2023-08-30
Adds an alias for nclass

eliocamp:


MichaelChirico:
Is this an exhaustive check of cases where `?S3generic` doesn't work?

eliocamp:
For reference, this is related to this bug report: https://bugs.r-project.org/show_bug.cgi?id=15272

PR #142, opened 2023-08-31
Bug 14359 (partial fix)

dickoa:
Tentative fix for

https://bugs.r-project.org/show_bug.cgi?id=14359

https://github.com/r-devel/r-project-sprint-2023/issues/33

### DIFF
--- a/src/library/stats/R/t.test.R
+++ b/src/library/stats/R/t.test.R
@@ -126,10 +126,12 @@ function(x, y = NULL, alternative = c("two.sided", "less", "greater"),
 }
 
 t.test.formula <-
-function (formula, data, subset, na.action, ...) 
+function (formula, data, subset, na.action = "na.pass", ...) 
 {
     if (missing(formula) || (length(formula) != 3L))
         stop("'formula' missing or incorrect")
+    if ("paired" %in% ...names())
+        stop("cannot use 'paired' in formula method")    
     oneSampleOrPaired <- FALSE
     if (length(attr(terms(formula[-2L]), "term.labels")) != 1L) 
         if (formula[[3L]] == 1L)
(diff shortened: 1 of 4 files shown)


mmaechler:
Thank you very much!

  All looks fine to me  but *not* you deleting the graphical display from the examples:
Examples are also about teaching.  Doing statistics in such a simple case without visualizing the problem nicely is  *_bad_* statistical practice; hence, the plot must remain.
But please, there is no need to update your patch!  We (R-core, possibly I) will review the patch anyway, and we'd rather not you clutter up bugzilla unnecessarily just because of such a small thing.  Hence I'm giving the feedback here instead of in bugzilla.
(Remark: I think it's waste of resources / unnecessary aggravation of the CO_2 situation to always run these expensive github actions; but no, I don't want to discuss this here; happy to do so during a coffee / lunch break.. ;-))

shannonpileggi:
Hi @mmaechler! thank you for the feedback. I am open to in person discussion if our paths cross today, but in case we do not, here are my thoughts.

I wholeheartedly agree with the sentiment that visualizations are part of good statistical practice. However, to me, the purpose of function documentation is to demonstrate implementation and execution of the function. The nuances surrounding good statistical practice, appropriate statistical methods, and pedagogical considerations seem out of scope for the examples section. I also suspect that when individuals run examples, they often do not execute or review all code in the examples section, but rather the bits that they feel are relevant to their aims.

shannonpileggi:
I am sorry I did not get the chance to discuss with you in person prior to my departure @mmaechler! 

I have considered this further, and if you feel passionately there is no harm in keeping the plot.

The corresponding figure for the new two sample t-test example would be 

``` 
plot(mpg ~ am, data = mtcars)
```

PR #149, opened 2023-12-19
partial match optional patch from Ivan Krylov

tdhock:
this is from https://stat.ethz.ch/pipermail/r-devel/2023-December/083073.html

### DIFF
--- a/src/library/base/R/dataframe.R
+++ b/src/library/base/R/dataframe.R
@@ -598,14 +598,14 @@ data.frame <-
 ###  These are a little less general than S
 
 `[.data.frame` <-
-    function(x, i, j, drop = if(missing(i)) TRUE else length(cols) == 1)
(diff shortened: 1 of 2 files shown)


src/library/base/R/dataframe.R
```diff
@@ -598,14 +598,14 @@ data.frame <-
 ###  These are a little less general than S
 
 `[.data.frame` <-
-    function(x, i, j, drop = if(missing(i)) TRUE else length(cols) == 1)
+    function(x, i, j, drop = if(missing(i)) TRUE else length(cols) == 1, pmatch.rows = TRUE)
```
MichaelChirico, line 601: Should there be an option like `warnPartialMatch{Args,Attr,Dollar}` to throw a warning if code is relying on the default behavior?

MichaelChirico: Just saw your comment later in the thread implying you won't pursue this as a patch :) resolving

PR #151, opened 2024-01-10
base: use a symlink for which instead of hard-coded path

haampie:
R hard-codes the path to the `which` executable inside bytecode, which is
stored in a compressed format.

This makes it very hard to relocate R itself at a later stage, or update the
path to `which` if it ever moves on the system.

This patch creates a symlink `rlib/R/library/base/R/which ->`
`/path/to/bin/which` so it's easier to relocate R / which, as you can just
modify a symlink instead of dealing with `base.rdb` and `base.rdx` files.

### DIFF
--- a/share/make/basepkg.mk
+++ b/share/make/basepkg.mk
@@ -72,15 +72,15 @@ mkRbase:
 	  else \
 	    cat $(RSRC) > "$${f}"; \
 	  fi; \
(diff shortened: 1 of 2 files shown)


haampie:
Looks like CI for windows is broken due to timeouts when installing miktex

> Installing CTAN packages
> Sorry, but: Timeout was reached

PR #155, opened 2024-03-02
Use single curl multi handle to re-use connections

jeroen:
This is a minimal proof of concept patch to test the performance implications of using a shared curl "multi-handle" to re-use http connections for multiple downloads to the same server (hence avoiding repeated TLS handshakes).

A more extensive patch, which includes cleaning up all connections when R quits, is available in #166

Quick benchmark:

```r
options(repos=c(CRAN='https://cloud.r-project.org'))
x <- available.packages()
pkgs <- row.names(x)[1:100]
system.time(download.packages(pkgs, tempdir(), available = x))
```

Before:

```
   user  system elapsed
 15.212   1.113  24.098
```

After:

```
   user  system elapsed
  0.327   0.521   1.661
```

### DIFF
--- a/src/modules/internet/libcurl.c
+++ b/src/modules/internet/libcurl.c
@@ -565,8 +565,8 @@ static void download_cleanup(void *data)
 	if (c->hnd && c->hnd[i])
 	    curl_easy_cleanup(c->hnd[i]);
     }
-    if (c->mhnd)
-	curl_multi_cleanup(c->mhnd);
+    //if (c->mhnd)
+	//curl_multi_cleanup(c->mhnd);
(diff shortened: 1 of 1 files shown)


sgsokol:
Nice update.
However, one point remains unclear for me: as `curl_multi_cleanup(c->mhnd)` is now commented out, how and when  `c->mhnd` will be cleaned up?

jeroen:
This is just to show the proof of concept. We would probably need to call `curl_multi_cleanup` when R exits.

jeroen:
https://blog.r-project.org/2024/12/02/faster-downloads/index.html

PR #157, opened 2024-03-23
Use pkgconf from Rtools44

jeroen:
Given that base R now [requires rtools44](https://github.com/r-devel/r-svn/commit/85be5dbfb6ac585ff41ffe1b9b251a81ae9572aa) to build, we can take advantage of pkgconf and remove some hardcoded flags. This makes the code more portable for other toolchains, including future versions of rtools.


jeroen:
Seems addressed in 721fe66c1bcf80ef836ac0a591a761f1d37e4228

PR #164, opened 2024-04-22
[Bugzilla#17289] Error in scan() for string NA when na.strings is empty

MichaelChirico:
https://bugs.r-project.org/show_bug.cgi?id=17289


MichaelChirico:
Fixed (see buganizer)

PR #165, opened 2024-04-22
[Bugzilla#18710] Turn on warnPartialMatch* by default

MichaelChirico:
Now sent to Bugzilla:

https://bugs.r-project.org/show_bug.cgi?id=18710


llrs:
Wow, this many and only in tests, not even inside the R code...
Note that there are some changes in white spaces (and no other change), for example tests/arith-true.R Line 186.

MichaelChirico:
>Wow, this many and only in tests, not even inside the R code...

That's expected IMO, `codetools::checkUsage(suppressPartialMatchArgs=FALSE)` is part of `R CMD check` --> partial match usage in package code is much likelier to be detected (since issues with `$` / `attr()` are much rarer to begin with):

https://github.com/r-devel/r-svn/blob/faa2ca57cc03c26c49a5de39db980268c6c4cc67/src/library/tools/R/QC.R#L4546

MichaelChirico:
Closed, e.g. r86849

PR #166, opened 2024-04-24
Use shared curl multi handle with cleanup on exit

jeroen:
This is a more extensive version of #155, with a hook to cleanup libcurl when R quits.


jeroen:
https://blog.r-project.org/2024/12/02/faster-downloads/index.html

PR #172, opened 2024-07-12
Bug 18674 tools::toTitleCase() 

shannonpileggi:
Bug 18674 tools::toTitleCase() incorrectly capitalizes conjunctions (e.g. 'and') when using suspensive hyphenation corrected.

Also fixed spacing in documentation, added two examples in documentation, and added two tests for the fix.

### DIFF
--- a/src/library/tools/R/utils.R
+++ b/src/library/tools/R/utils.R
@@ -2688,6 +2688,8 @@ function(text)
         ## do not remove capitalization immediately after ": " or "- "
         ind <- grep("[-:]$", xx); ind <- ind[ind + 2L <= length(l)]
         ind <- ind[(xx[ind + 1L] == " ") & grepl("^['[:alnum:]]", xx[ind + 2L])]
+        # don't capitalize lpat words after hyphenation
+        ind <- ind[!(xx[ind] == "-" & grepl(lpat, xx[ind + 2L]))]
         l[ind + 2L] <- FALSE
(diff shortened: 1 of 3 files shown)


shannonpileggi:
fyi @sarahzeller @reikookamoto @Bisaloo

mmaechler:
Committed to R's sources now, w/ even more examples (+ NEWS + `tools-Ex.Rout.save`).
Thank you very much!

NB (also to myself!):  

1. Appending `.diff` to the end of the URL automatically provides something you *save* from your browser as plain text file, named, e.g.,  `toTitleCase_PR-18674.diff`   *and* then 
2. in R's bugzilla, you'd typically use the __Add an Attachment__ option to add that `*.diff` file as attachment:
       - on the new  _Create New Attachment_  web page, add a  __Brief Decription__ ,
       - also click the **Content Type:** __patch__ box,   and then
       - add comments (often prose text) below  rather than in the original bugzilla page for the PR.
       - press __Submit__
 
This will give you a perfect bugzilla submission, with only *one* e-mail to all of R-core plus PR author and more listeners.

MichaelChirico:
Linking here for convenience: https://bugs.r-project.org/show_bug.cgi?id=18674

shannonpileggi:
Thank you @mmaechler for the detailed instructions and the feedback! I have added an issue to the contribution to guide to assess if we could use more information there.

https://github.com/r-devel/rdevguide/issues/177

PR #175, opened 2024-07-12
Fix max print bug 15027

Abhi-1U:
Alternative Implementation to fix bug https://bugs.r-project.org/show_bug.cgi?id=15027, https://github.com/r-devel/r-dev-day/issues/4

Test cases were originally written by @loreabad6.

### DIFF
--- a/src/main/printarray.c
+++ b/src/main/printarray.c
@@ -1,3 +1,4 @@
+
 /*
  *  R : A Computer Language for Statistical Data Analysis
  *  Copyright (C) 1995, 1996	Robert Gentleman and Ross Ihaka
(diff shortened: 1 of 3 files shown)


src/main/printarray.c
```diff
@@ -348,35 +349,63 @@ void printMatrix(SEXP x, int offset, SEXP dim, int quote, int right,
     if(c > 0 && R_print.max / c < r) /* avoid integer overflow */
 	/* using floor(), not ceil(), since 'c' could be huge: */
 	r_pr = R_print.max / c;
+    int c_pr = c;
+    if (c > R_print.max) {
+        c_pr = R_print.max;
+	r_pr = 1;
```
lionel-: I think this should remain set to 0 if `r` is 0?

tests/print-tests.R
```diff
@@ -339,5 +339,22 @@ as.complex(1:10)
...
+print(matrix(nrow = 100, ncol = 4), max = 5) ## works
+## fix: ommiting rows **and** columns and warning printed
+## case: omiting rows and columns
+print(matrix(nrow = 10, ncol = 4), max = 3)
+## case: omitting rows
+print(matrix(nrow = 10, ncol = 2), max = 5)
+## case: omitting cols, at least one row prints
+print(matrix(nrow = 1, ncol = 6), max = 5)
+## in R 4.4.0 there should be a warning for ommited rows (NEW BUG)
+print(array(dim = c(2, 4, 1)), max = 3)
```
lionel-, line 354: There was a typo in the issue, this is actually:

```r
print(array(dim= c(2, 2, 1)), max = 2)
```

src/main/printarray.c
Abhi-1U: Thanks for spotting the mistake. 
so adding a check for `r > 0` before setting `r_pr` to `1`  should be a good fix.

PR #176, opened 2024-07-13
Use `DATAPTR_OR_NULL()` in `ExtractSubset()` when possible

DavisVaughan:
https://github.com/r-devel/r-dev-day/issues/23
https://bugs.r-project.org/show_bug.cgi?id=18766

### DIFF
--- a/src/main/subset.c
+++ b/src/main/subset.c
@@ -66,31 +66,42 @@ static R_INLINE SEXP VECTOR_ELT_FIX_NAMED(SEXP y, R_xlen_t i) {
    currently the subscript code forces allocation.
 */
 
-#define EXTRACT_SUBSET_LOOP(STDCODE, NACODE) do { \
-	if (TYPEOF(indx) == INTSXP) {		  \
-	    const int *pindx = INTEGER_RO(indx);  \
-	    for (i = 0; i < n; i++) {		  \
-		ii = pindx[i];			  \
-		if (0 < ii && ii <= nx) {	  \
-		    ii--;			  \
-		    STDCODE;			  \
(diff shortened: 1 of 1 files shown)


src/main/subset.c
```diff
@@ -134,36 +145,52 @@ attribute_hidden SEXP ExtractSubset(SEXP x, SEXP indx, SEXP call)
     PROTECT(result = allocVector(mode, n));
     switch(mode) {
     case LGLSXP:
-	EXTRACT_SUBSET_LOOP(LOGICAL0(result)[i] = LOGICAL_ELT(x, ii),
-			    LOGICAL0(result)[i] = NA_INTEGER);
+	EXTRACT_SUBSET_LOOP(const int* px = (const int*) DATAPTR_OR_NULL(x),
+	                    LOGICAL0(result)[i] = px[ii],
+	                    LOGICAL0(result)[i] = LOGICAL_ELT(x, ii),
+			    LOGICAL0(result)[i] = NA_LOGICAL);
```
krlmlr, line 151: I wonder if something along the following lines would work too, with a suitable implementation above:

```suggestion
	EXTRACT_SUBSET_LOOP(const int* px = (const int*) DATAPTR_OR_NULL(x),
	                    SET_LOGICAL_ELT, LOGICAL_ELT, NA_LOGICAL);
```

`SET_LOGICAL_ELT()` could be defined as a macro, like in `src/main/saveload.c` .

krlmlr (commented):
Nice! Spaces vs. tabs isn't super-clean everywhere, but let's wait for the Bugzilla feedback once it's there.

PR #177, opened 2024-08-17
[Bugzilla#17672] [Bugzilla#18782] Allow seq.{Date,POSIXt} without specifying from=

MichaelChirico:
cc @shannonpileggi

Closes https://bugs.r-project.org/show_bug.cgi?id=18782 and https://bugs.r-project.org/show_bug.cgi?id=17672.


MichaelChirico:
Fixed e.g. r87284

PR #179, opened 2024-08-23
[Bugzilla#18786] make showMethods() display a syntactic name

MichaelChirico:
https://bugs.r-project.org/show_bug.cgi?id=18786


MichaelChirico:
Fixed: c87048

PR #188, opened 2024-11-20
Update checkS3methods to ignore deprecated/defunct functions - R Dev Day issue 76

luciorq:
Context: [r-devel/r-dev-day#76](https://github.com/r-devel/r-dev-day/issues/76)

Also related: [r-devel/cran-cookbook/discussions/53](https://github.com/r-devel/cran-cookbook/discussions/53)

This Pull Request introduces two new auxiliary functions, `isDefunct()` and `isDeprecated()`, and also integrates them into `checkS3methods()` to handle deprecated and defunct methods.

The development of this feature was carried out under the guidance of @gmbecker and @lawremi.

### DIFF
--- a/src/library/tools/R/QC.R
+++ b/src/library/tools/R/QC.R
@@ -2387,7 +2387,7 @@ checkTopLevelCall <- function(expr, fun_name) {
     fun_name <- as.name(fun_name)
     fun_called <- expr[[1]]
     if (is.call(fun_called)) {
-        inner_called <- fun_called[[1]
+        inner_called <- fun_called[[1]]
         if (identical(inner_called, quote(`:::`)) ||
             identical(inner_called, quote(`::`))) {
            fun_called <- fun_called[[3]]
--- a/src/library/tools/R/QC.R
+++ b/src/library/tools/R/QC.R
@@ -2635,6 +2635,11 @@ function(package, dir, lib.loc = NULL)
                   function(g) {
                       methods <-
                           gen_dot_cls_matches(g, functions_in_code)
+                      method_funs <-
+                          Filter(Negate(function(x) {
+                              isDefunct(x) || isDeprecated(x)
+                              }), mget(methods, code_env))
+                      methods <- names(method_funs)
                       if((n <- length(methods)) > 0L) {
                           gargs <- nfg(g, code_env)
                           entries <-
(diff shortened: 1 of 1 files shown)


src/library/tools/R/QC.R
```diff
@@ -2379,6 +2379,38 @@ function(x, ...)
...
+### Additional functions for checkS3methods
+checkTopLevelCall <- function(expr, fun_name) {
+  if (inherits(expr, "if") || !is.call(expr)) {
+    return(FALSE)
+  }
+  fun_name <- as.name(fun_name)
+  fun_called <- expr[[1]]
+  if (is.call(fun_called)) {
+    inner_called <- fun_called[[1]] 
+    if (as.character(inner_called) %in% c(":::", "::")) {
```
aitap: `as.character` on calls is dangerous combined with an `if`. Consider `expr <- quote( cmocean::cmocean('haline')(n) )`. Now `fun_called <- expr[[1]]` is `quote( cmocean::cmocean('haline') )` and `inner_called <- fun_called[[1]]` is `quote( cmocean::cmocean )`, so `as.character(inner_called)` is `c("::", "cmocean", "cmocean")` ("the condition has length > 1").

It's safer but more verbose to test this with ``identical(inner_called, quote(`:::`)) || identical(inner_called, quote(`::`))``.

<del>I think that `as.character(inner_called)[[1]] %in% c(":::", "::")` should also work because there shouldn't be any 0-element calls.</del> <ins>(No, that tests for the wrong condition and would break below.)</ins>

luciorq: Thank you for pointing that out. Your suggestion is implemented as of 
8e71230.

luciorq: Your suggestion is implemented as of 
e0f6a85

luciorq:
Closing as this is already solved.

 [r-devel/r-dev-day#76](https://github.com/r-devel/r-dev-day/issues/76#issuecomment-2592724935)
 
 > Kurt committed a variant of this patch in revision [SVN r87570](https://github.com/r-devel/r-svn/commit/4ce8bac8f0c4eca481519cccf7a47c6bcd4960cc) with a [small tweak](https://github.com/r-devel/r-svn/commit/da0280a0955870f605b89c0dd2f6a5ed3546fadd) today. He says:
> 
> > I just committed a variant of the patch with c87570.  Unfortunately, I
> > managed to commit too much in one commit, so the commit msg does not
> > give credit etc.  Sorry.
> 
> > The change looks a bit different as I re-used and enhanced exisiting
> > code, ensured to match PKG::.Deprecated and PKG::.Defunct when PKG is
> > base but not otherwise, and changed to filter at the end as ideally we
> > do get() calls to get the functions from their names as little as
> > possible.
>

PR #196, opened 2025-02-24
Lockdir prefix for install packages

zeehio:
Some distributed file systems do not support opening files in append mode. These file systems are often used in data analysis cloud platforms.

R package installation relies on appending to files, for instance collating R code or when installing help pages.

Therefore, packages can't be installed in those filesystems. Instead, users are forced to install packages into a local directory and copy them afterwards.

However, the current package installation procedure already uses a 00LOCK directory to install packages there, before copying them to the final library directory.

By globally modifying the location of the 00LOCK directory, it is possible to use a local filesystem to install packages, where append is allowed. The installation process takes care of copying the resulting package into the final out directory.

This change introduces the environment variable PKG_LOCKDIR_PREFIX that, when set to a directory like "/tmp/r-lockdir", uses that root path to create all 00LOCK folders.

This allows to install packages on file systems that do not support opening files in append mode.

### DIFF
--- a/doc/NEWS.Rd
+++ b/doc/NEWS.Rd
@@ -423,6 +423,15 @@
 
       \item There is preliminary support for C++26 with GCC \eqn{>=} 14, Apple
       \command{clang++} \eqn{>=} 16 and \I{LLVM} \command{clang++} \eqn{>=} 17.
+
+      \item \code{install.packages()} supports setting a different path for the
+	  intermediate \code{00LOCK} directories through the environment variable
+	  \code{R_LOCKDIR_PREFIX}. This helps source package installation in some
+	  distributed filesystems available in cloud environments like databricks.
+	  Check the
+	  \dQuote{R Installation and Administration} manual for further details on
+	  this option.
+
     }
   }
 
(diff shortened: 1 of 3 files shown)


llrs:
This repository is not official to submit an enhancement I recommend to do so on the official page: https://bugs.r-project.org/

See [this guide](https://contributor.r-project.org/rdevguide/chapters/submitting_feature_requests.html) on how to do so. On the guide there is also a page on how to request the account to be able to submit to the R core members this patch.

I would recommend to add some details to the proposal like under which circumstances the file system do not accept appending. Maybe there is a workaround and the r-devel mailing list could help you.

zeehio:
Thanks Lluís! I Saw the guidelines and I will eventually submit a proper email asking for feedback and giving context. I wanted to make sure my PR was working as expected.

The context is when I try to install pkgs on a databricks volume, which lives on a distributed file system. I have workarounds but I feel this solution would be simpler.

Still, I realized yesterday that my patch should work well with relative paths and currently it does not, so I wanted to fix that before requesting comments to the r-devel mailing list

PR #197, opened 2025-03-04
[Bugzilla#18865] Use any(.) over sum(.) > 0

MichaelChirico:
https://bugs.r-project.org/show_bug.cgi?id=18865

Double-checking I didn't do anything sloppy

### DIFF
--- a/src/library/methods/R/RClassUtils.R
+++ b/src/library/methods/R/RClassUtils.R
@@ -2412,7 +2412,7 @@ classesToAM <- function(classes, includeSubclasses = FALSE,
       maybe <- if(!is.null(f <- get0(x, where))) is.function(f) else FALSE
       if(maybe)
         maybe <- is(f, "genericFunction") ||
-              (length(grep("UseMethod", deparse(f))) > 0) ||
+              any(grepl("UseMethod", deparse(f))) ||
               is.primitive(f)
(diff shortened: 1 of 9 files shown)


llrs (commented):
I like those patches!! Thanks!
Just a minor change to make it slightly faster (I thought this was the purpose of these changes, feel free to ignore for a faster review from R core)

src/library/methods/R/RClassUtils.R
```diff
@@ -2412,7 +2412,7 @@ classesToAM <- function(classes, includeSubclasses = FALSE,
       maybe <- if(!is.null(f <- get0(x, where))) is.function(f) else FALSE
       if(maybe)
         maybe <- is(f, "genericFunction") ||
-              (length(grep("UseMethod", deparse(f))) > 0) ||
+              any(grepl("UseMethod", deparse(f))) ||
```
llrs, line 2415:
```suggestion
              any(grepl("UseMethod", deparse(f), fixed = TRUE)) ||
```

MichaelChirico: good catch! Agree it's an improvement. I don't have an easy workflow for updating patches, so I'll just make a note of it in the issue tracker.

PR #200, opened 2025-05-22
✨ Add SPDX field to the license.db

ThierryO:
This is task 1 of r-devel/r-dev-day#84 @llrs

### DIFF
--- a/share/licenses/license.db
+++ b/share/licenses/license.db
@@ -7,6 +7,7 @@ FSF: free (https://www.gnu.org/licenses/license-list.html#GPLv2)
 File: share/licenses/GPL-2
 URL: https://www.r-project.org/Licenses/GPL-2
 FOSS: yes
+SPDX: GPL-2.0-only
 
 Name: GNU General Public License
 Abbrev: GPL
@@ -18,16 +19,18 @@ File: share/licenses/GPL-3
 URL: https://www.r-project.org/Licenses/GPL-3
 FOSS: yes
 Extensible: yes
+SPDX: GPL-3.0-only
 
 Name: GNU Library General Public License
 Abbrev: LGPL
 Version: 2
 SSS: LGPL-2
-OSI: open 
+OSI: open
 FSF: free
 File: share/licenses/LGPL-2
 URL: https://www.r-project.org/Licenses/LGPL-2
 FOSS: yes
+SPDX: LGPL-2.0-only
 
 Name: GNU Lesser General Public License
 Abbrev: LGPL
@@ -38,6 +41,7 @@ FSF: free (https://www.gnu.org/licenses/license-list.html#LGPLv2.1)
 File: share/licenses/LGPL-2.1
 URL: https://www.r-project.org/Licenses/LGPL-2.1
 FOSS: yes
+SPDX: LGPL-2.1-only
 
 Name: GNU Lesser General Public License
 Abbrev: LGPL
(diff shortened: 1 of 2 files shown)


share/licenses/license.db
```diff
@@ -18,16 +19,18 @@ File: share/licenses/GPL-3
...
 FOSS: yes
 Extensible: yes
+SPDX: GPL-3.0-only
 
 Name: GNU Library General Public License
 Abbrev: LGPL
 Version: 2
 SSS: LGPL-2
-OSI: open 
+OSI: open
```
llrs-roche: Noticing the whitespace change:

```suggestion
OSI: open (https://opensource.org/license/lgpl-2-0)
```

llrs-roche (approved):
Looks good. I think we can proceed to task 2

On a related note did you verify that OSI information here and on SPDX match? Probably it hasn't changed but there could be some differences. Not that I expect them but it might help identify some inaccuracies.

ThierryO:
I'm not sure how the check the OSI field. SPDX has only "isOsiApproved" which is either true or false.

llrs-roche:
Yes, that's what I meant yes is what on this file is open. But that's not important. Let's see how we do task2 that is the difficult part.

src/library/tools/R/license.R
```diff
@@ -611,6 +614,8 @@ function(x)
                                stringsAsFactors = FALSE)
         }
 
+        sapply(expansions, `[[`, "SPDX") |>
```
llrs-roche: Some R developers try to avoid `sapply` because the simplify parameter can break further steps. Not sure if it is better to use `lapply` and then `unlist`.

llrs-roche (commented):
The second step looks very minimal. I can't test this how it works at the moment (will try to do that later today). 

One thing that needs to be done before the R core accept this is document the new argument & functionality on the corresponding .Rd file. But I can do it once I test this.

src/library/tools/R/license.R
ThierryO: Given that the SPDX field will always be a string, we could replace it with
```suggestion
        vapply(expansions, FUN.VALUE = character(1), FUN = function(x) {x[["SPDX"]]}) |>
```

llrs:
I had some troubles checking this (building R from git repo didn't work well on my machine) and then I found Kurt had already patched R.
The `sapply` line ended up like:

https://github.com/r-devel/r-svn/blob/a9fcda6ec62283fcf3750f58d2202017d93ab1b1/src/library/tools/R/license.R#L617-L619

Congrats on the patch!

eddelbuettel:
> building R from git repo didn't work well on my machine

@ThierryO said the same in Mons. I should look into that again.  My risk avoidance strategy of ... sticking with svn worked on the 'travel' laptop as it does here (and has for umpteen years) on the 'home' desktop/server.

ThierryO:
> > building R from git repo didn't work well on my machine
> 
> @ThierryO said the same in Mons. I should look into that again. My risk avoidance strategy of ... sticking with svn worked on the 'travel' laptop as it does here (and has for umpteen years) on the 'home' desktop/server.

I got it working now (on Ubuntu 24.04) with the script below

```
cd full_path_to_repo_root
sed -i.bak 's|$(GIT) svn info|./.github/scripts/svn-info.sh|' Makefile.in
./.github/scripts/wget-recommended.sh
./.github/scripts/svn-info.sh
CC=gcc
./configure --enable-R-shlib --with-blas --with-lapack --disable-java --prefix=/usr/local/lib/R-devel
make
```

eddelbuettel:
@ThierryO That looks right. I have done it before too, and of course CI does it all the time.  Thanks for documenting!

llrs:
Thanks for the script @ThierryO I simply created a patch file and applied to the svn repo and continued there. 
I asked Kurt about task 3 as we were talking about license issues on Bioconductor.

PR #205, opened 2025-08-02
#96 , improvement in error message for rep() function of R , 

Siddd-hrr:
Made the errorcall() more informative to understand exactly where the things are going wrong during  initialization of  "times" attributes  of rep() function .

### DIFF
--- a/src/main/subassign.c
+++ b/src/main/subassign.c
@@ -156,7 +156,7 @@ static SEXP EnlargeVector(SEXP x, R_xlen_t newlen)
 	warning(_("assignment outside vector/list limits (extending from %lld to %lld)"),
 		(long long)len, (long long)newlen);
 
-    /* if the vector is not shared, is growable. and has room, then
+    /* if the vector is not shared, is growable, and has room, then
        increase its length */
     if (! MAYBE_SHARED(x) &&
 	IS_GROWABLE(x) &&


src/main/seq.c
```diff
@@ -745,7 +745,9 @@ attribute_hidden SEXP do_rep(SEXP call, SEXP op, SEXP args, SEXP rho)
 	    len = lx * it * each;
 	} else { // nt != 1
 	    if(nt != (double) lx * each)
-		errorcall(call, _("invalid '%s' argument"), "times");
+		    errorcall(call,_("invalid '%s' argument: length of time = %d, expected = %d from (length of x = %d , each = %d) "),"times",nt,(int)((double) lx * each),(int) lx , (int) each );
```
aitap: At least some of the variables being cast to `(int)` are of type `R_xlen_t`. Could they hold a number exceeding `INT_MAX`?

src/main/errors.c
```diff
@@ -3006,3 +3006,40 @@ void R_InitConditions(void)
...
+
+attribute_hidden
+SEXP R_makeCpuTimeLimitError(SEXP call, double cpu_time)
+{
+    return R_makeTimeLimitError(call,
+                                cpu_time,
+                                "cpu_TimeLimitError",
+                                "cpu");
+} 
+
```
eddelbuettel: I am not sure how much by R Core you have looked at but I do not think they would be thrilled with this indentation default....

eddelbuettel:
You don't need to close it, but if you want to submit it should look more like their code -- as a basic courtesy.

Siddd-hrr:
> You don't need to close it, but if you want to submit it should look more like their code -- as a basic courtesy.

The pull request was raised by mistake , actually I am still working on it but , instead  of just stagging my changes I also pushed which led to raise of Pull request for totally different Uid issue no. ,  and also thank you for your advice  I will consider it will working on any issue.

PR #206, opened 2025-08-11
Implement missing binding accessors and constructors

DavisVaughan:
With @lionel- 
Checked in https://github.com/r-lib/rlang/pull/1818


lionel-:
Updated patch that includes second commit: [env-accessors.patch](https://github.com/user-attachments/files/25795568/env-accessors.patch)

PR #222, opened 2025-12-28
bitmask approach for mayHaveNaNOrInf()

JanMarvin:
Originally I went with a bitmask approach. This looked promising in benchmarks. But build into R it was even slower.
Next try: use fabs() < DBL_MAX. Looking at the performance:  _\*sigh\*_


JanMarvin:
This should be ~4 times faster than the previous `mayHaveNaNOrInf()` and should speed up cases where #221 is not possible

JanMarvin:
There is no use in this. `option("matprod" = "blas")` seems to be the simpler solution.

PR #223, opened 2026-01-01
Documentation fixes for paste.Rd and substr.Rd

aakritixyz:
This PR includes minor documentation improvements to base R functions:
	1.	paste.Rd
	•	Clarified why paste0() is slightly more efficient than paste().
	•	Explained recycle0 behavior when collapse is specified.
	2.	substr.Rd
	•	Added a note about zero-length extraction: when start or stop results in zero-length, an empty string "" is returned.
	•	Updated examples to demonstrate zero-length behavior.

These changes improve clarity for users reading the documentation, without modifying any code behavior.

### DIFF
--- a/src/library/base/man/paste.Rd
+++ b/src/library/base/man/paste.Rd
@@ -140,12 +140,5 @@ paste(    collapse = "|")
 paste(    collapse = "|", recycle0 = TRUE)
 paste({}, collapse = "|")
 paste({}, collapse = "|", recycle0 = TRUE)
-
-## Demonstrating recycle0 behavior with collapse argument
-vals_empty <- character(0)
-vals_nonempty <- c("a", "b")
-
-paste(vals_empty, vals_nonempty, collapse = ",")       # -> "a,b"  (empty vector ignored)
-paste(vals_empty, vals_nonempty, collapse = ",", recycle0 = TRUE) # -> "" (length 1, empty)
 }
 \keyword{character}
--- a/src/library/base/man/paste.Rd
+++ b/src/library/base/man/paste.Rd
@@ -46,7 +46,8 @@ paste0(\dots,            collapse = NULL, recycle0 = FALSE)
   desirable, e.g.\sspace{}in \code{paste("the value of p is ", p)}.
 
   \code{paste0(\dots, collapse)} is equivalent to
-  \code{paste(\dots, sep = "", collapse)}, slightly more efficiently.
+  \code{paste(\dots, sep = "", collapse)}, slightly more efficiently
+  as it avoids handling a separator.
 
   If a value is specified for \code{collapse}, the values in the result
   are then concatenated into a single string, with the elements being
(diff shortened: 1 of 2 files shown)


JesseAlderliesten:
The bullet-point list mentions 'Explained recycle0 behavior when collapse is specified.' for paste.Rd but as far as I can tell, you did not change the explanation about recycle0?

JesseAlderliesten:
The examples added to substr.Rd are useful, but the comment '## Demonstrating zero-length extraction behavior' is misleading: 'character(0)' in the first example indeed is zero-length, but '""' in the second example has length 1, even though it contains zero characters: length("") == 1 but nchar("") == 0.
Maybe rephrase it as '## Extraction behavior from zero-length strings and empty strings'?

aakritixyz:
Hi Sir

Thanks for your feedback! I apologize for not including the changes you suggested in my previous commit , I realized after pushing that I had only partially staged the updates, so the examples and comments weren’t fully updated.

I’ve now incorporated your suggestions:
	•	In substr.Rd, the comment has been clarified to distinguish zero-length vectors (character(0)) from empty strings ("").
	•	In paste.Rd, examples now demonstrate recycle0 behavior when collapse is specified.

Both changes are included in the latest commit. PTAL , thanks!

JesseAlderliesten:
No problem! However, I would not add the examples in paste.Rd, since there are already examples showing the use of 'collapse' and 'recycle0' (below the headings ## 'recycle0 = TRUE' and '## When 'collapse = <string>').
If you want to keep the examples, your first example should probably use paste0(...) instead of paste(...) because paste(...) returns " a, b".

aakritixyz:
Thanks for the clarification!
I’ve removed the additional examples from paste.Rd to avoid redundancy, since the use of collapse and recycle0 is already well covered under the existing sections.

No new examples are added now, and the documentation remains unchanged functionally. Please let me know if this looks good or if you’d prefer a different approach.

JesseAlderliesten:
Looks good to me, but you'll have to wait for a response of one of the R maintainers. I'm just a fellow-coder who saw your remark in the R-contributers Slack.

aakritixyz:
Thanks for taking a look and for the helpful guidance!
I’ll wait for feedback from one of the R maintainers. Much appreciated.

llrs:
To get feedback from R maintainers the patch should be submitted to either r-devel.
I saw your [message on December 31st](https://stat.ethz.ch/pipermail/r-devel/2025-December/084317.html), but the patch attached didn't reach my inbox, perhaps the readers didn't see it.

In any case, I would submit to bugzilla as there it has less chances of being forgotten.

aakritixyz:
@llrs That makes sense regarding visibility and the usual workflow for R. For now, I mainly wanted to get initial feedback and ensure the documentation changes were aligned with expectations, so I started with a GitHub PR.

Based on the guidance here, I’ll follow up via the appropriate channel (r-devel / Bugzilla) if further action is needed. Thanks again for the helpful advice and for pointing me in the right direction.

eddelbuettel:
> but the patch attached didn't reach my inbox,

AFAIK mailman is configured to strip attachments. Bugzilla is clearly preferred by the R Core team.

PR #224, opened 2026-01-02
[Bugzilla #18988] [i18n] Unify similar messages for translation in R-base.pot

MichaelChirico:
https://bugs.r-project.org/show_bug.cgi?id=18988

Testing patch to be submitted, to make sure there's no tests assuming certain messages, and that I didn't make any silly mistakes.


llrs (commented):
I particularly like the changes where a common pattern is extracted. :+1:

Did you/AI check if there are similar strings with `dQuote` or `sQuote`? Probably these functions can't be used everywhere but your comment/changes about "" and '' reminded me of this.

MichaelChirico:
> I particularly like the changes where a common pattern is extracted. 👍
> 
> Did you/AI check if there are similar strings with `dQuote` or `sQuote`? Probably these functions can't be used everywhere but your comment/changes about "" and '' reminded me of this.

Could you write an example of what you specifically have in mind? want to be sure I'm understanding. I didn't specifically prompt around that (this time), though it could be added. 

one potential issue is that the plain .pot files don't have the code context, so we'd either want to let an agent selectively read source files (difficult because R source locations are not currently copied into the .pot file), or build potools into the loop (which does add the string context to the data frame it builds)

llrs:
Sometimes messages are built with:

```r
stop("A word ", dQuote(variable), " more.") 
stop("A word ", sQuote(variable), " more.")
```

And sometimes they use quotes:

```r
stop(sprintf('A word "%s" more.', variable))
stop(sprintf("A word '%s' more.", variable))
```

If `getOption("useFancyQuotes") == FALSE` the second error messages are equivalent to the first one.

I am wondering if there are strings that are duplicated with only this kind of differences where translating one (and be more consistent with `dQuote()` and `sQuote()` on error messages) would be worth. 

But given Martin Mächler comment I'm no longer sure that adding more scope to that issue is a good idea.

PR #225, opened 2026-01-03
Fix merge.data.frame to preserve column names when merging empty data frames (Bug 18930)

aakritixyz:
When merging empty data frames using merge() (for example with all.x = TRUE and/or all.y = TRUE), the resulting data frame loses its column names or shows incorrect names. This behavior is inconsistent and can break code that expects stable column names — especially in workflows that programmatically inspect or manipulate results from merge().

This addresses Bug #18930 in the R Bugzilla tracker: “Inconsistent behaviour of merge() on empty DF”, where column naming depends on arbitrary internal formatting rather than the input frame’s names.

This patch updates merge.data.frame so that when one or both of the data frames are empty, the function still preserves all column names in the result while maintaining correct behavior for non‑empty merges. No other merge logic is changed.

**Test-case:**
merge(data.frame(x = numeric()), data.frame(y = 1:2),
      all.x = TRUE, all.y = TRUE)
#>   x y
#> <0 rows> (or 0‑length row.names)

merge(data.frame(x = numeric(), y = numeric()),
      data.frame(z = 1:2),
      all.x = TRUE, all.y = TRUE)
#>   x y z
#> <0 rows> (or 0‑length row.names)

This change only affects behavior when merging truly empty data frames. All existing behavior for non‑empty input remains unchanged.

### DIFF
--- a/src/library/base/R/merge.R
+++ b/src/library/base/R/merge.R
@@ -1,186 +1,73 @@
-#  File src/library/base/R/merge.R
-#  Part of the R package, https://www.R-project.org
-#
-#  Copyright (C) 1995-2018 The R Core Team
-#
-#  This program is free software; you can redistribute it and/or modify
-#  it under the terms of the GNU General Public License as published by
-#  the Free Software Foundation; either version 2 of the License, or
-#  (at your option) any later version.
-#
-#  This program is distributed in the hope that it will be useful,
-#  but WITHOUT ANY WARRANTY; without even the implied warranty of
-#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#  GNU General Public License for more details.
-#
-#  A copy of the GNU General Public License is available at
-#  https://www.R-project.org/Licenses/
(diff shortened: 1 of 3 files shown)


aakritixyz:
Hi R Core team,

This patch addresses Bug #18930: inconsistent behavior of merge() when merging empty data frames. Specifically, it ensures that column names are preserved when one or both data frames are empty, without affecting the behavior for non-empty merges.


I have verified the fix locally on macOS:
merge(data.frame(x = numeric()), data.frame(y = 1:2), all.x = TRUE, all.y = TRUE)
#> x y
#> <0 rows> (or 0-length row.names)

merge(data.frame(x = numeric(), y = numeric()), data.frame(z = 1:2), all.x = TRUE, all.y = TRUE)
#> x y z
#> <0 rows> (or 0-length row.names)

However, the CI builds are failing on Linux (Debian, ARM/Intel) and Windows, which seems related to cross-platform build requirements rather than the logic of the patch itself.

Could you please advise on whether any adjustments are needed to make this patch compatible across all target platforms?


Regards,
Aakriti

@bbolker

PR #228, opened 2026-01-21
Add attribute to versions

llrs:
Test adding an attribute to keep the character on version numbers


llrs:
Reported on Bugzilla: https://bugs.r-project.org/show_bug.cgi?id=19004

PR #231, opened 2026-02-08
Added na.rm for Summary.numeric_version

marberts:
https://bugs.r-project.org/show_bug.cgi?id=19000

### DIFF
--- a/src/library/base/R/version.R
+++ b/src/library/base/R/version.R
@@ -37,125 +37,144 @@
 ## module on CPAN.)
 
 .make_numeric_version <-
-function(x, strict = TRUE, regexp, classes = NULL)
-{
+  function(x, strict = TRUE, regexp, classes = NULL) {
     ## Internal creator for numeric version objects.
 
     nms <- names(x)
 
-    if(!length(x)) {
-        y <- list()
+    if (!length(x)) {
+      y <- list()
     } else {
-        if(!is.character(x)) {
-            msg <- gettextf("invalid non-character version specification 'x' (type: %s)",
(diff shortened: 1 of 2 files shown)


llrs (commented):
I'm sure this will help with the `summary()` but how will it help with `max()`, `min()`, `range()`? They will be called independently of the `summary.numeric_version()` method.

marberts:
Thanks for taking a look @llrs. This is for the `Summary.numeric_version()` method for the group generic (capital S), not `summary.numeric_version()`. (There actually isn't a method for `summary()` for `numeric_version` objects.) Each of `min()`, `max()`, and `range()` dispatch to `Summary.numeric_version()` when passed a `numeric_version` object, and are the only such functions that don't immediately throw an error for not being applicable.

Hopefully this clarifies the patch, but please let me know if the patch/bugzilla report aren't clear so that I can fix it.

llrs:
Hi, I am interested on improving support for `package_version` and saw your enhancement (my PR is on #228 but somehow it breaks the CI).

Now that I think of it it would be nice if a test is added on `tests/reg-test-1e.R` to make sure this works as intended. 
I didn't follow up how `max()`, `min()`, `range()` are dispatched internally to `Summary` But if the tests pass then it will be easier to merge (and I'll be reassured that it happens without checking too deep the code :).

llrs:
Great! One last things is to directly attach the diff to bugzilla as it might make easier for R core members to pick that than check the link and copy the patch from the web.

marberts:
Good idea. I've updated the bugzilla.

PR #233, opened 2026-02-19
feat: Allow ALTREP row names unmodified

krlmlr:


tests/reg-tests-2.Rout.save
```diff
@@ -5853,8 +5853,7 @@ data frame with 0 columns and 4 rows
 structure(list(X1.4 = 1:4), class = "data.frame", row.names = c(NA, 
 -4L))
 > dput(d1)
-structure(list(X1.4 = 1:4), class = "data.frame", row.names = c(NA, 
-4L))
+structure(list(X1.4 = 1:4), class = "data.frame", row.names = 1:4)
```
DavisVaughan, line 5856: From reading this test, the reason the row.names are now `1:4` in the `dput` output for `d1` is because they were created with

```r
d1 <- data.frame(1:4, row.names=1:4)
```

Internally this does

```r
row.names(value) <- row.names
# i.e. row.names(value) <- 1:4
```

and because `1:4` is already ALTREP, `row_names_gets` no longer tries to convert it to the `c(NA, -4L)` compact version. I think this is a good thing!

PR #237, opened 2026-03-02
Fix `debugonce()` and `undebug()` issues with primitive generics

lionel-:
This patch fixes two bugs:

1. `debugonce()` on a primitive generic like `[` does not respect "once" semantics. It debugs every subsequent S3 dispatch indefinitely.

2. `undebug()` cannot cancel a pending `debugonce()` on any function type, including closures:

```r
> debugonce(identity)
> undebug(identity)
Warning message:
In undebug(identity) : argument is not being debugged
> identity(1)
debugging in: identity(1)
debug: x
Browse[1]>
```

R uses two flags for debugging: `RDEBUG` (set by `debug()`) and `RSTEP` (set by `debugonce()`). For closures, `debugonce()` works because `applyClosure` checks `RSTEP`, enters the debugger, and clears the flag. The next call sees `RSTEP = 0`.

Primitive generics (primitives that perform S3 dispatch, such as `[`, `[[`, `$`) never go through `applyClosure`. They call `dispatchMethod()` via `usemethod()` directly. `dispatchMethod()` propagates the debug state to the method but never clears `RSTEP` on the primitive, so it persists across all subsequent dispatches.

The `undebug()` bug is simpler: it only cleared `RDEBUG`, not `RSTEP`. For closures this went unnoticed probably because the "once" clearing in `applyClosure` works. After one call the flag is gone anyway.

Fixes:

- `dispatchMethod()`: after propagating debug state from a primitive generic to the method, clear `RSTEP` on the primitive.

- `undebug()`: clear both `RDEBUG` and `RSTEP`, and check both flags for the "not being debugged" warning.

### DIFF
--- a/src/main/debug.c
+++ b/src/main/debug.c
@@ -49,9 +49,10 @@ attribute_hidden SEXP do_debug(SEXP call, SEXP op, SEXP args, SEXP rho)
 	SET_RDEBUG(CAR(args), 1);
 	break;
     case 1: // undebug()
-	if( RDEBUG(CAR(args)) != 1 )
+	if( RDEBUG(CAR(args)) != 1 && RSTEP(CAR(args)) != 1 )
 	    warning("argument is not being debugged");
 	SET_RDEBUG(CAR(args), 0);
+	SET_RSTEP(CAR(args), 0);
 	break;
     case 2: // isdebugged()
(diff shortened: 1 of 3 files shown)


lionel-:
@DavisVaughan Would you like to take a look before I send this to the bugzilla?

PR #239, opened 2026-03-20
Don't look up through parents in C dots API

lionel-:
Follow up for https://bugs.r-project.org/show_bug.cgi?id=18928


src/main/envir.c
```diff
@@ -1577,9 +1577,22 @@ static int ddVal(SEXP symbol)
 
 #define length_DOTS(_v_) (TYPEOF(_v_) == DOTSXP ? length(_v_) : 0)
 
+/* Walk parent environments to find the first one containing `...` */
+SEXP R_findDots(SEXP env)
```
DavisVaughan:
```suggestion
SEXP R_findDotsEnv(SEXP env)
```

Otherwise this reads as if it returns the DOTSXP!

src/main/names.c
```diff
@@ -931,6 +931,9 @@ FUNTAB R_FunTab[] =
 {"dotDelayedExpression", do_dotDelayedExpr, 0, 11, 2,  {PP_FUNCALL, PREC_FN,	0}},
 {"dotDelayedEnvironment", do_dotDelayedEnv, 0, 11, 2,  {PP_FUNCALL, PREC_FN,	0}},
 {"dotForcedExpression", do_dotForcedExpr, 0, 11, 2,    {PP_FUNCALL, PREC_FN,	0}},
+{"dotsLength", do_CDotsLength,		0, 11,	1,      {PP_FUNCALL, PREC_FN,	0}},
+{"dotsNames", do_CDotsNames,		0, 11,	1,      {PP_FUNCALL, PREC_FN,	0}},
+{"dotsElt", do_CDotsElt,		0, 11,	2,      {PP_FUNCALL, PREC_FN,	0}},
```
DavisVaughan, line 936: is this supposed to be aligned in some way?

DavisVaughan (approved):
I like the exhaustive set of tests, and they make sense to me

PR #242, opened 2026-03-26
Ensure `.pkg.type()` returns something when type isn't "source"

VisruthSK:


s-u:
@VisruthSK by pushing this branch you made me aware of the issue, so thanks! I did solve directly in R to avoid the overhead of pushing things across systems, but I just wanted to thank you.

VisruthSK:
Thanks for fixing that so quickly! Yeah going through bugzilla and all that would've taken forever, so I'm very glad you happened to see this PR.

PR #249, opened 2026-04-03
Rbuildignore more dev files by default

lionel-:
- Removes a local `.hidden_file_exclusions` list in `check.R` that masked the global one in `utils.R`, so we don't have to maintain two lists.

- Added exclusion patterns for common development files that should not be shipped in packages:
  - IDE/tooling: `.Rproj`, `.Rproj.user`, `.vscode`, `.cache`, `.editorconfig`, `revdep`, `CRAN-SUBMISSION`
 
  - Agents: `.claude`, `.codex`, `.gemini`
  - C/C++ tooling: `.clangd`, `compile_commands.json` (nested matching, these may appear in subdirectories). Also added `TAGS` to the existing GNU Global pattern.
  - CI/docs: `codecov.yml`, `_pkgdown.yml`, `.github`, `LICENSE.md`, `CODE_OF_CONDUCT.md`, `cran-comments.md`
  - R tooling: `air.toml` / `.air.toml` (nested matching as these support per-directory config), `.lintr`
  - Auth files: `.httr-oauth`, `.Renviron`, `.Rprofile` (these are a security concern if they get bundled and sent away)


src/library/tools/R/utils.R
```diff
@@ -575,22 +575,35 @@ function(file, pdf = FALSE, clean = FALSE, quiet = TRUE,
...
 .hidden_file_exclusions <-
-    c(".Renviron", ".Rprofile", ".Rproj.user",
+    c(## R
+      ".Renviron", ".Rprofile",
       ".Rhistory", ".Rapp.history",
-      ".tex", ".log", ".aux", ".pdf", ".png",
-      ".backups", ".cvsignore", ".cproject", ".directory",
-      ".dropbox", ".exrc", ".gdb.history",
+      ## Build artifacts
+      ".tex", ".log", ".aux", ".pdf", ".png", ".cache",
```
eddelbuettel, line 583: How about adding `.bbl`, `.blg` which `bibtex` may leave behind?

eddelbuettel (approved):
Looks good, and a nice cleanup.  One minor suggestion inline.

src/library/tools/R/utils.R
lionel-: Done!

eddelbuettel: Nice, thank you.  We still need to add those to `cleanup` whereever it happens but in case that did not run may as well preclude these from the generated archive.

lionel-:
https://bugs.r-project.org/show_bug.cgi?id=19045

novica:
`jarl.toml` should aslo be added to this list.

PR #250, opened 2026-04-08
Implement `...(arg)` forwarding syntax

lionel-:
POC for https://github.com/RConsortium/S7/issues/592

This adds a `...(arg)` syntax that forwards a named argument (here `arg`) across layers of function calls using the same mechanism as `...`. This allows wrapping NSE functions that inspect their arguments with `substitute()`, and have expression capture work transparently across the wrapper:

```r
label <- function(x) deparse(substitute(x))

with_forward <- function(value) label(...(value))
without_forward <- function(value) label(value)

with_forward(sin(1:10))
#> [1] "sin(1:10)"

without_forward(sin(1:10))
#> [1] "value"
```

For instance, this makes it possible to wrap `plot()` and have the internal labeling of dimensions work as expected:

```r
# Wrapper that changes some arguments
bare_plot <- function(x, y) plot(...(x), ...(y), axes = FALSE)

# labels "sin(1:10)" and "cos(1:10)", not "x" and "y"
bare_plot(sin(1:10), cos(1:10))
```

As a bonus, this syntax can be a full base R replacement for `{{` in tidyeval, without any changes to rlang. The capture side `enquo()` is implemented on top of the new environment accessors, and would be immediately compatible with arguments forwarded with `...()`, capturing both the relevant expression and its environment.

Finally, that mechanism would also solve the forwarding issue for S7, which is the motivating case for this POC:

```r
gen <- function(x, y, ...) {
  method <- find_method(class_of(x))
  method(...(x), ...(y), ...)
}
```

### DIFF
--- a/src/include/Defn.h
+++ b/src/include/Defn.h
@@ -2256,6 +2256,7 @@ void process_site_Renviron(void);
 void process_system_Renviron(void);
 void process_user_Renviron(void);
 SEXP promiseArgs(SEXP, SEXP);
+SEXP dotsForwardSym(SEXP);
 int Rcons_vprintf(const char *, va_list);
 int REvprintf_internal(const char *, va_list);
 SEXP R_data_class(SEXP , Rboolean);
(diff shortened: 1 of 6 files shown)


t-kalinowski:
This is an elegant solution!

PR #251, opened 2026-04-15
Bugzilla: 7195; R-intro changes

llrs:
[Bug 7195](https://bugs.r-project.org/show_bug.cgi?id=7195)

An update to the patch provided as plain text.
Code output has been checked on R 4.5.3

I only omitted a chunk that I don't think is no longer relevant.


llrs:
Patch submitted with r-intro.R changes

PR #252, opened 2026-04-21
Don't serialise growable bit

lionel-:
https://github.com/r-lib/vctrs/issues/2144

https://bugs.r-project.org/show_bug.cgi?id=19056


jeroen:
Can you rebase on main because sadly you started from a broken commit in the svn.

jeroen:
Ah seems it is already merged :)

PR #253, opened 2026-04-25
Fix build for distros that use --package-metadata

jeroen:
bug


jeroen:
Merged in https://github.com/r-devel/r-svn/commit/145221c96f2a7d0ee8dc53cd5ff5ae7fd4c7f28e

PR #257, opened 2026-05-23
Bug 18743

llrs:
[Bug 18743](https://bugs.r-project.org/show_bug.cgi?id=18743).

Update the documentation to show the "forward" and "backward" approach for distance and the ward.D* methods


llrs:
Closed with a different patch

PR #262, opened 2026-06-10
Make trace() work for functions whose S3 class has multiple strings

hadley:
trace() failed for any function with a multi-string S3 class attribute, e.g. S7 generics and methods (class `c("S7_method", "function", "S7_object"))`, with  `'length = 3' in coercion to 'logical(1)'` because the class vector was passed where a single class name was expected. Reported via RConsortium/S7#584.

In `methods:::.TraceWithMethods`:

* Use `.class1()` rather than `class()` to derive the `"<class>WithTrace"` class name, so the trace class extends the (S4 view of) the most specific class instead of `setClass()` receiving a vector of names.

In `methods:::.initTraceable`:

* Use `.class1()` likewise for the old class lookup.

* Skip the `as(.Object, oldClass) <- def` slot copy when oldClass is virtual, i.e. an S3 class registered by setOldClass(). Such classes have no slots to copy beyond .Data, and for them the coercion replaced .Object wholesale with def, losing the trace wrapper.

* Extract the data part with `slot(def, ".Data")` instead of `def@.Data`: since R 4.3, `@` on non-S4 objects dispatches on S3 methods, so a package method such as S7's `@.S7_object` intercepted the call and errored.

Tracing an S7 *method* now works end to end with no changes to S7. 

Tracing an S7 *generic* additionally needs S7's C-level change (RConsortium/S7#695): the traced binding is an S4 object of class `"S7_genericWithTrace"` whose link to `"S7_generic"` exists only in the S4 class table, so the `Rf_inherits(generic, "S7_generic")` check in S7's `method-dispatch.c` cannot see it and S7 must unwrap the "original" attribute itself. That is inherent to how `trace()` wraps functions and not fixable in R: an S4 object's class attribute is its S4 class, and making `inherits()` consult S4 inheritance would be far more invasive. S7's R-level workaround (pre-registering the WithTrace classes) remains useful on R versions without this fix and for property access and printing of traced generics.

Adds a regression test in `tests/classes-methods.R` and a `NEWS` entry.


lawremi:
The change to avoid calling `as<-()` when the class is virtual was just guarding against a deeper issue with how methods was defining replace functions for old class extensions. It is possible to have virtual old classes with slots and non-virtual old classes that would still fail. I'm working on a fix for this.

Related, S7 should probably use the `S4Class=` arg to `setOldClass()` to give the  `S7_method` old class the slots that represent its S7 class properties.

Also, I'm not sure about using `slot<-()` over `@<-()`, because in principle `@<-()` needs to work on these objects with `.Data`; it's a contract any override of `@<-()` must adhere to. S7 adheres to it after the S4 extension branch is merged. As an aside, `@()` hits the S4 path before dispatch, so it avoids this issue. We should probably make the two consistent at some point.

For now I'm taking the changes to use `class1()` forward into R.

lawremi:
Update: the change to use `.class1()` and the fix for `as(.Object, oldClass) <- def` have been committed to R. The final issue around `@<-` will be resolved via the S4 inheritance branch.

hadley:
@lawremi I updated this branch just in case it's useful you to see the diff.

PR #263, opened 2026-06-11
Add object() and make deparse() produce evaluable code for bare OBJSXPs

hadley:
Fixes the problem reported in RConsortium/S7#524: `dump()`, `dput()`, and `deparse()` of an `OBJSXP` without the S4 bit set (the structure underlying every S7 object) emitted the unparseable literal `<object>`, silently dropping all attributes.

## Changes

- **New base function `object()`** returning a "bare" object of type `"object"`: no attributes, no S4 bit. Implemented as a `.Internal` wrapping the existing `R_allocObject()` allocator (already used by the serializer, which is why `readRDS()` round-trips worked while deparsing did not).
  - `src/main/objects.c`: `do_object()`, next to `do_setS4Object()`
  - `src/main/names.c`, `src/include/Internal.h`: FUNTAB entry and declaration
  - `src/library/base/R/methodsSupport.R`: R-level wrapper, next to `asS3()`/`asS4()`
  - `src/library/base/man/object.Rd`: documented as internal/expert-use
- **`deparse()` support**: the `OBJSXP` case in `deparse2buff()` now emits `object()` wrapped via the standard `structure(...)` attribute machinery (`attr1()`/`attr2()`, following the `CLOSXP` pattern), implementing the commented-out sketch that was already in that branch. The result is no longer flagged unsourceable, so the "deparse may be incomplete" warning is gone.
- Docs (`dump.Rd` note, NEWS) and a regression test in `tests/reg-tests-1e.R` asserting the deparse -> parse -> eval round-trip is `identical()`.

## Notes

This does *not* technically make a new SEXPTYPE available from R: a bare `OBJSXP` could already be created without C code via `asS3(getClass("S4")@prototype, complete = FALSE)` (Luke Tierney's trick, see the S7 issue). `object()` is just a base-level spelling of the same thing that needs neither the methods package nor the S4 class cache, making it suitable as deparse output: dumped objects now evaluate with only base attached, e.g.

    structure(object(), class = "S7_object")

is `identical()` to `S7::S7_object()` without S7 installed.

Verified against a full build: new regression test passes, full `make test-Reg` clean, real S7 instances round-trip through deparse/eval (`all.equal()` modulo closure environments, deparse's documented limitation for functions), S4 deparsing and printing unchanged.


t-kalinowski:
We might want to expand the signature to take `...`, which could become attributes. (There are already commended out sections in the `dput()` code path for this kind of usage). 

Usage could then be:

```r
object(foo = bar)
```

instead of 
```r
structure(object(), foo = bar)
```

lawremi:
I haven't tried it, but looking that the code, would `is.object(object())` return `FALSE` because no object bit (class attribute) has been set? That would be weird. In other words, should `R_allocObject()` set the object bit?

t-kalinowski:
Reconciling `is.object()` with `OBJSXP` is going to be difficult/impossible. They mean fundamentally different things. But maybe we can reduce confusion by requiring that the `object()` constructor require a `class` argument? E.g.:

```r
object <- function(class, ...) {
  if(missing(class)) stop("class must be supplied")
  obj <- R_allocObject()
  attributes(obj) <- list(class = class, ...)
  obj
}
```

hadley:
I like that proposal.

PR #264, opened 2026-06-12
Disallow `|` commands in `postscript()` (and hence `pdf()`)

hadley:
`file` is, in the vast majority of cases, a path — inert data naming where bytes should land. But when its first character is `|`, the remainder is handed to `R_popen()`, i.e. executed by `/bin/sh -c`. So `pdf(file = x)` is a write-to-disk operation for almost all values of `x` and an arbitrary-command-execution operation for a small, invisible subset of them. Nothing at the call site signals that the second mode exists, and it requires a close read of the documentation to discover it exists.

It's reasonable to expect that a careful R programmer will know that `system()` is potentially dangerous when given unsanitized user inputs; I don't think they have any particular reason to suspect that `pdf()` is just as bad. This matters because filenames are routinely derived from untrusted input. R is now deployed in Shiny, plumber, OpenCPU and batch report-generation services where it's reasonable for the filename to be a user-specified parameter.

The cost of removing this functionality is very small because the legitimate capability of piping device output to a program remains fully available through `postscript(file = "", command = …)` and `print.it = TRUE`.

### DIFF
--- a/doc/NEWS.Rd
+++ b/doc/NEWS.Rd
@@ -8,7 +8,12 @@
 \section{\Rlogo Changes in R-devel}{
   \subsection{Significant User-Visible Changes}{
     \itemize{
-      \item .
+      \item \code{pdf()} and \code{postscript()} no longer interpret a
+      \code{file} argument of the form \code{"|cmd"} as a command to
+      pipe the output to, and now signal an error: this was a
+      potential security risk.  To print PostScript output, use
+      \code{postscript(file = "", command = ...)} or
+      \code{print.it = TRUE}.
     }
   }
 
--- a/src/library/grDevices/R/unix/dev2bitmap.R
+++ b/src/library/grDevices/R/unix/dev2bitmap.R
@@ -77,14 +77,14 @@ bitmap <- function(file, type = "png16m", height = 7, width = 7, res = 72,
     extra <- ""
     if (!is.na(taa)) extra <- paste0(" -dTextAlphaBits=", taa)
     if (!is.na(gaa)) extra <- paste0(extra, " -dGraphicsAlphaBits=", gaa)
-    cmd <- paste0("|", shQuote(gsexe),
+    cmd <- paste0(shQuote(gsexe),
                   " -dNOPAUSE -dBATCH -q -sDEVICE=", type,
                   " -r", res,
                   " -dAutoRotatePages=/None",
                   " -g", ceiling(res*width), "x", ceiling(res*height),
                   extra,
                   " -sOutputFile=", shQuote(file), " -")
-    postscript(file = cmd, width = width, height = height,
+    postscript(file = "", command = cmd, width = width, height = height,
                pointsize = pointsize, paper = "special", horizontal = FALSE, ...)
     invisible()
 }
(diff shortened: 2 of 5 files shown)


MichaelChirico:
FWIW it looks very rare in checked-in code, so few people are ostensibly relying on this "feature":

https://github.com/search?q=%2F%5B%5E%5Cw.%5D%28pdf%7Cpostscript%29%5C%28%5Cs*%5B%27%22%5D%5B%7C%5D%2F+-path%3Apostscript.Rd+-path%3Apostscript.R+-path%3Apdf.Rd+-lang%3Ayaml+-lang%3Ahtml+-path%3Asvn-base&type=code

hadley:
@MichaelChirico interesting that they're all `pdf(| cat)`

s-u:
I'd like to point out that any app that does not pass input through a regexp filter is insecure by definition. Even other characters that are conceptually valid in paths like like either kind of slash are dangerous, so all apps must validate input and pipe should be one of the banned characters.

In principle I'm not against reducing the attack surface, but I was somewhat surprised how hard it is to replace `pdf("|cat")` - mainly because `pdf` doesn't support connections (to a point where it creates files named after the connection ID silently, which seem suboptimal). So I'd say an ideal solution would be to add support for connections and then it could be simply `pdf(stdout())` or the general `pdf(pipe("blah","w"))` for all other use-cases.

hadley:
@s-u solving that in general is a bit tricky because the PDF device currently requires a seekable sink:

- 26 `ftell(pd->pdffp)` sites record each object's absolute byte offset for the
  cross-reference (`xref`) table written at the end (`devPS.c:8145-8157`).
- `rewind()` to patch the header after the fact (`devPS.c:8159-8161`).
- `fseek()` for stream-length measurement during compression (`devPS.c:8229-8231`).

And obviously the `pipe()` and `stdout()` connections are not seekable.

One simple approach would be to write the whole PDF to a seekable temp file, then stream that temp
file to the connection. Does that approach seem reasonable to you?

s-u:
@hadley thanks for looking. Yeah, that's annoying. Especially in the context of services I'd prefer ways to avoid creating files which is why `pdf("|cat")` is nice - I guess `rawConnection()` would work, but that's getting further off field.

As I have now commented on the PR it may be better to just make the interpretation of the pipe optional based on anew argument, and just error out if the pipe is used when not allowed (default) but if allowed use the current behavior.

TroyHernandez:
Side effect worth noting: this diff also removes an error-path resource leak.

`PDF_endfile()` calls `fclose(pd->pdffp)` and `pclose(pd->pipefp)` only after the copy loop, so an error during the `fwrite()` longjmps past both. On Linux `|true` triggers it via the SIGPIPE handler.

```
for (i in 1:40) {
    pdf(file = "|true")
    plot(1:3)
    try(dev.off(), silent = TRUE)
}
```

r90391: 40 failures leave 80 open fds and 40 unreaped children, linearly; 40 successful calls to a file are flat. (`/proc/self/fd` and `ps`; LSan tracks neither.)

Deleting the block fixes it. Keeping the path behind an opt-in argument means it needs unwind-safe cleanup.

PR #268, opened 2026-06-16
Junk commit to break R if bugzilla 19086 test fails

kaskr:


jeroen:
Thank you for the contribution Kasper. Could you maybe update the description above and add a summary description of the bug and patch, and a link to the bugzilla issue? This makes it easier to review.

kaskr:
Actually it's already merged here: f4b70c1b8e36b57707027aec2e672792d7cf53cc
So closing now.

PR #270, opened 2026-06-26
Bug 18842: warn when length(col) != length(levels)-1 in filled.contour(); updated docs

darakhshannehal:
Fixes [Bug 18842](https://bugs.r-project.org/show_bug.cgi?id=18842): `filled.contour()` silently recycles or drops colors when `length(col) != length(levels) - 1`, instead of warning. This is easy to trigger accidentally since `nlevels` doesn't directly determine `length(levels)` (it goes through `pretty()`, which can return a different count).

Adds a `warning()` reporting the actual vs. expected color count, and clarifies the `col`/`levels`/`nlevels` relationship in the docs. 

Tested against the reprexes in the bug report, warns on mismatch, silent on correct/default calls.

Related: r-devel/r-dev-day#144

@pmur002  flagging since you're on the originating issue.

### DIFF
--- a/src/library/graphics/R/filled.contour.R
+++ b/src/library/graphics/R/filled.contour.R
@@ -48,10 +48,14 @@ function (x = seq(0, 1, length.out = nrow(z)),
         y <- x$y
         x <- x$x
     }
-    if (any(diff(x) <= 0) || any(diff(y) <= 0))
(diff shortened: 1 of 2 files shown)


pmur002:
Thanks @darakhshannehal . Patch committed to r-devel (r90197)

PR #271, opened 2026-06-26
enforce UTF-8 encoding for DCF read/write

kevinushey:
This implements [r-devel/r-dev-day#156](https://github.com/r-devel/r-dev-day/issues/156) (proposed by @kurthornik, building on #86): bring R's DCF read/write functions into compliance with the Debian Control File requirement that control files be encoded in UTF-8.

### Changes

- **`read.dcf()` now returns UTF-8** in both code paths:
  - `all = TRUE` (R, `src/library/base/R/dcf.R`): reads as UTF-8 and repairs invalid byte sequences with `iconv(sub = "byte")`; drops the old `Encoding <- "unknown"` marking.
  - `all = FALSE` (C, `src/main/dcf.c`, `do_readDCF`): new helpers `utf8_char_len()` (a streaming single-character validator adapted from `valid_utf8()`) and `mkCharUTF8sub()`, which escape invalid bytes as `<xx>` and mark the result UTF-8. Fast-path (no copy) when the string is already valid.
- **`write.dcf()` always writes UTF-8** regardless of the session locale: values with a declared encoding are converted, invalid bytes escaped, and the output written byte-verbatim. The `useBytes` argument is now ignored (kept for back-compatibility, documented as such).
- **`tools:::.read_description()` / `.write_description()`** no longer special-case the `Encoding` field.
- Docs (`dcf.Rd`), a NEWS entry, and regression tests (`reg-tests-1e.R`).

Invalid bytes are escaped (`<xx>`), not dropped, so the C and R paths produce identical output.

### Notable behavior change

The `Encoding` field of a `DESCRIPTION` is no longer used to re-encode the other fields. A non-UTF-8 (e.g. `Encoding: latin1`) `DESCRIPTION` now reads with its high bytes escaped rather than transcoded. This is the intended effect of enforcing UTF-8 (issue Task 2) and is noted in NEWS. Repository index generation (`PACKAGES` / `PACKAGES.gz`) is now UTF-8 regardless of the build locale.

### Testing

Built R trunk and verified in both `LC_ALL=C` and a UTF-8 locale. The new `reg-tests-1e.R` block covers: C/R-path agreement; 2-, 3-, and 4-byte valid characters; invalid sequences (truncated lead byte, overlong, UTF-16 surrogate, isolated byte); literal `<ff>` not double-escaped; invalid byte in a field name; repeated-field list columns; continuation-line escaping; latin1 -> UTF-8 on write; and the `keep.white` (non-folding) write path. The regression suite (`reg-tests-1*`, `reg-IO`, `reg-IO2`, `reg-packages`, `reg-encodings`, `reg-examples*`, `reg-S4`, `test-Iconv`) and the `base`/`tools`/`utils` examples all pass.

### DIFF
--- a/doc/NEWS.Rd
+++ b/doc/NEWS.Rd
@@ -8,7 +8,18 @@
 \section{\Rlogo Changes in R-devel}{
   \subsection{Significant User-Visible Changes}{
     \itemize{
-      \item .
+      \item \code{read.dcf()} and \code{write.dcf()} now treat \abbr{DCF}
+      files (such as package \file{DESCRIPTION} and repository
+      \file{PACKAGES} files) as \abbr{UTF-8}, as required by the Debian
+      control file format.  \code{read.dcf()} returns values in
+      \abbr{UTF-8} and \code{write.dcf()} always writes \abbr{UTF-8}
+      regardless of the session locale; bytes which are not valid
+      \abbr{UTF-8} are preserved by escaping them as \samp{<xx>} rather
+      than being re-encoded via a \samp{Encoding} field.  As a
+      consequence, the \code{useBytes} argument of \code{write.dcf()} is
+      ignored, and non-\abbr{UTF-8} \file{DESCRIPTION} files (e.g. those
+      declaring \samp{Encoding: latin1}) are no longer re-encoded on
+      reading.
     }
   }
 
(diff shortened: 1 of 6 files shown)


jeroen:
Your PR needs to target branch `main` instead of `trunk` for the CI to run.

kevinushey:
Ah, I'll fix that up -- thanks!

PR #208, opened 2025-08-30
WIP: PR for 18880

llrs:
[The enhancement aims](https://bugs.r-project.org/show_bug.cgi?id=18880) to make it easier to find existing functions via links on the help page.
See also some suggestion on a [thread in fosstodon](https://fosstodon.org/@Lluis_Revilla/115102629612645497).

PR #274, opened 2026-07-09
Link from `conflicts` to `conflictRules` and vice versa in `See Also`

JesseAlderliesten:


llrs:
This is a great addition! Is it related to any feature request on bugzilla? I have a "draft" PR with similar proposals on #208. Do you mind if I take this and incorporate it there? I was planning on submitting it soon to test waters on changes like these.

JesseAlderliesten:
Dear Lluís,

It is in Bugzilla [#19099](https://bugs.r-project.org/show_bug.cgi?id=19099). There, Martin Maechler replied these links will it will be incorporated, and sollicited for working examples to be incorporated in addition to the links.

I intend to work on providing such examples, and for that will first create such examples for argument `warn.conflicts` of `library()` and `require()`, and for option `conflicts.policy` which influences the behaviour of that argument. That makes this PR more involved than 'just' adding cross links. So I prefer to keep this separate, even though I agree the added linking itself would fit in your proposals in #208 .

llrs:
No problem, it's better some continuos small progress, than no big progress at all. 

I see that different R core people have different preferences. But I too saw that some documentation files lacked examples and was considering adding them in addition of links. So your patch was already very helpful. Thanks

PR #287, opened 2026-08-03
[Bugzilla#19117/c1] Change order in which mbcstowcs error is checked

MichaelChirico:
https://bugs.r-project.org/show_bug.cgi?id=19117#c1


MichaelChirico:
r90409

PR #289, opened 2026-08-04
[Bugzilla#19122] Validate %OS within 0-60

MichaelChirico:
https://bugs.r-project.org/show_bug.cgi?id=19122


MichaelChirico:
r90409

PR #307, opened 2026-08-24
RFC: prototype int64 and uint64 as opaque ALTSXP classes

kevinushey:
## Summary

This is an alternative prototype to #303. It starts from the implementation and tests in that proposal, but replaces the dedicated `XINTSXP` representation with:

- an opaque atomic `ALTSXP` SEXPTYPE;
- built-in `int64` and `uint64` ALTREP classes;
- class-supplied representation methods (`Element_size`, `Elt`, `Set_elt`, and `Get_region`); and
- semantic ALTREP methods for `Binary_op`, `Unary_op`, `Compare`, `Hash`, `Coerce`, `Format`, `Summary`, and `Combine`.

`typeof()` reports `"alt"`, while the registered ALTREP class and `storage.mode()` identify `int64` or `uint64`. Generic data-pointer access to the opaque outer object is refused. The built-in classes use a native-endian raw vector as private `data1` storage.

Binary operations and comparisons use two-sided negotiation: the left opaque class is asked first, followed by a distinct right class if the left declines. Operand order is preserved and a separate dispatch argument identifies the receiver.

The base implementation routes arithmetic, relational operations, coercion, formatting and printing, summaries, hashing, and concatenation through these methods where appropriate. Serialization uses portable ALTREP class state in the existing version-3 format.

The package-facing probe defines unrelated opaque ALTREP classes and verifies all proposed semantic hooks, including right-hand fallback, same-class single dispatch, ordinary numeric promotion, implicit class identity, default printing, and protection from the built-in integer private-layout fallback.

## Scope and limitations

- This prototype intentionally supports only signed and unsigned 64-bit integers.
- Hash collision resolution requires `Elt` to return a canonical representation for semantically equal elements.
- Package-defined opaque classes still need existing ALTREP facilities, or additional semantic hooks, for operations outside this proposal such as arbitrary class-defined sorting and subsetting.
- As with any new SEXPTYPE, previously exhaustive package `TYPEOF()` switches will reach their default case. The prototype makes this explicit rather than masquerading as `REALSXP`.

## Verification

Passed locally on macOS arm64:

- full R build;
- imported #303 int64/uint64 regression and development suites;
- C interface and package-defined ALTSXP dispatch suite;
- base `xinteger.R` and `storage-info.R` tests;
- Specific tests;
- Reg tests; and
- Internet tests.

The umbrella `make check` progressed through the package examples until `tcltk`, where this local configuration had selected x86_64 Tcl/Tk libraries for an arm64 build and `tclVersion()` was unavailable. All preceding examples passed; the Tcl/Tk failure is unrelated to this patch.


kevinushey:
Follow-up pushed in eb1d9ee3e3: the xinteger prototype no longer uses `sxpinfo.gp` bits. Width and signedness are properties of the built-in ALTREP class (`altint64` versus `altuint64`), the raw payload remains in `data1`, and the per-instance reserved-NA policy is stored as a scalar logical in `data2`. Serialized class state now carries only the endian-neutral payload and NA policy; kind comes from class identity. The focused FFI and full XInts suites pass.

kevinushey:
Closing; https://github.com/r-devel/r-svn/pull/308 is a cleaner start.

PR #312, opened 2026-08-28
Use wide Windows APIs for locale handling

kevinushey:
## Summary

- Use `_wsetlocale()` for `Sys.getlocale()` and `Sys.setlocale()` on Windows.
- Translate the R locale argument to UTF-16 once with `wtransChar()` and reuse it across locale categories.
- Return `Sys.getlocale()` results as UTF-8 on every platform, converting native locale names with `reEnc()` outside Windows.
- Keep the implementation localized to `src/main/platform.c` through small `R_`-prefixed platform macros.

This supersedes #311, which retained the narrow Windows locale APIs.

## Regression coverage

The Windows-only regression test directly uses the canonical `Norwegian Bokmål_Norway.utf8` locale name, spelling the non-ASCII character as `\u00e5` in the source. It verifies that both locale APIs return UTF-8 strings, then switches `LC_CTYPE` to `C` and confirms that the declared UTF-8 locale name can still be passed through `Sys.setlocale()` and read back unchanged.

The checks remain conditional because locale availability depends on the Windows installation.

## Validation

- `git diff --check`
- Compiled `src/main/platform.c` successfully in a freshly configured out-of-tree macOS build (`make -C src/main platform.o`).
- Parsed and ran `tests/reg-encodings.R` successfully on macOS; the Windows-only regression block is skipped there.

## Follow-up

A further commit addresses review findings:

- On POSIX, `R_LOCALE_ARG` was `CHAR`, so with `Sys.getlocale()` now returning UTF-8 a non-ASCII name would no longer round-trip through `Sys.setlocale()` in a non-UTF-8 locale. It is now `translateChar()`, the counterpart of `wtransChar()`; both return either the CHARSXP's own data or `R_alloc()`ed memory, which is what makes hoisting the translated name above the `switch` safe.
- `R_LOCALE_MKCHAR` performs the native-to-UTF-8 conversion on POSIX, so `Sys.setlocale()` returns the same encoding as `Sys.getlocale()` there too, and `do_getlocale()` loses its remaining `#ifdef Win32`. ASCII names are unaffected: `mkCharLenCE()` does not mark them, so `Encoding()` still reports `"unknown"` for them.
- The regression test runs in every Windows session, not only UTF-8 ones -- a non-UTF-8 session with a `.utf8` locale name is the case the UTF-16 conversion helps most -- and now also covers `"LC_ALL"`, the one branch with its own sequence of category calls and a re-read of the compound result.
- NEWS entry, and a sentence in `?Sys.getlocale` on the UTF-8 return value.

Validated on macOS against a full build of the branch: `tests/reg-encodings.R` passes, and `Sys.setlocale(cat, Sys.getlocale(cat))` round-trips for every category, with ASCII names returned unmarked as before. The Windows block still relies on CI.

### DIFF
--- a/doc/NEWS.Rd
+++ b/doc/NEWS.Rd
@@ -294,6 +294,17 @@
 
       \item Fix for \PR{19035} (piping \code{postscript()} output)
       thanks to \I{Jeroen Ooms} and \I{Ivan Krylov}.
+
+      \item \code{Sys.setlocale()} on Windows now passes the locale name
+      to the C runtime as UTF-16, so that a name containing non-ASCII
+      characters -- such as \code{"Norwegian Bokmål_Norway.utf8"} or
+      \code{"Turkish_Türkiye.utf8"}, the default names on recent Windows
+      versions -- can be set from any locale, including \code{"C"}.
+      Previously the name was passed as bytes to be interpreted in the
+      \emph{current} locale, so that, e.g., restoring a locale saved by
+      \code{Sys.getlocale()} after switching to \code{"C"} failed.  On
+      all platforms, \code{Sys.getlocale()} and \code{Sys.setlocale()}
+      now return non-ASCII locale names marked as \abbr{UTF-8}.
     }
   }
 }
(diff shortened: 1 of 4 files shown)


kevinushey:
Temporarily closing and reopening to trigger CI after retargeting this PR from trunk to main.

jeroen:
Can you rebase this on main, because the trunk was broken at the time of this fork :/

kevinushey:
Done!

PR #313, opened 2026-09-01
Detect binary package format by magic bytes when unpacking

jeroen:
Follow-up to the custom-binaries work (7cd9a82a76): both binary unpackers currently assume the archive format from context rather than from the file itself.

- `unpackPkgZip()` (win.binary path) always runs unzip, so a custom binary tarball served from a `bin/windows/...` repository cannot be unpacked there.
- The `untar` wrapper in `.install.binary()` passes everything to `utils::untar()`. Zip files (as currently served by r-universe for `windows.binary.clang-aarch64`) only extract by accident when `untar()` falls back to an external bsdtar (e.g. `tar.exe` from System32), which auto-detects zip. With `TAR=internal` or a GNU tar this fails.

This patch makes both sides sniff the 4-byte zip signature (`PK\x03\x04`) and dispatch to `unzip()` or `utils::untar()` accordingly, keeping the existing `getOption("unzip")` handling for zips in `unpackPkgZip()`. `untar()` already detects gzip/bzip2/xz/zstd tarballs by magic bytes itself, so with this change the archive format of a binary repository no longer depends on platform conventions or the local tar implementation. This also benefits `R CMD INSTALL` on Windows, which routes `*.zip` files through `unpackPkgZip()`.

Tested: both files parse; the dispatch extracts a zip and a tar.gz correctly, and an empty/truncated file is not misdetected as zip (falls through to the usual `untar()` error).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/library/utils/R/install.bin.R
+++ b/src/library/utils/R/install.bin.R
@@ -96,7 +96,15 @@
     }
 
     ## not sure why the above was used - possibly it pre-dates utils::untar()?
-    untar <- function(what, where) utils::untar(what, exdir=where)
+    ## binary packages can be tarballs or zip files: dispatch on the
+    ## magic bytes as the file extension is not reliable
+    untar <- function(what, where) {
+        magic <- readBin(what, "raw", n = 4L)
+        if (identical(magic, as.raw(c(0x50, 0x4b, 0x03, 0x04))))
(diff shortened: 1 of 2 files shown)


s-u:
Thanks! In R now as r90464

PR #320, opened 2026-09-02
grow environment hash tables at 0.75 bindings per chain, to prime sizes that suit the PJW hash

kevinushey:
## Summary

Hashed environments grow their table at a measured load factor instead of an inconsistent count, to prime sizes chosen for the PJW hash. Filling an environment with a million bindings one at a time is 2.7x faster, lookups in large environments visit 1.2 bindings instead of 1.75, and `env.profile()` reports the number of non-empty chains correctly.

Stacked on #316 (the cache in `main` still shares the bucket count and `R_HashSizeCheck()` with environment tables in five places which that PR rewrites); this PR is the last commit.

## What was wrong

- `R_HashSet()` increments the count for every new binding: its search loop always ends at an empty chain, so the `if (ISNULL(chain))` guard never failed. A resize, `attach()` and `R_RestoreHashCount()` (after loading a saved environment) count occupied chains instead. So `env.profile()$nchains` was wrong after any insert, and the resize threshold, 85% of that mixed count with 20% growth, was neither an occupancy nor a load: measured tables sat at 1.0 to 1.2 bindings per chain, 1.4 to 1.75 bindings visited per hit, and filling an environment to a million bindings moved each binding 5.7 times.
- Each binding visited on a chain is a cache miss on a cons cell somewhere in the heap, so the load factor sets the cost of a lookup, as for the CHARSXP cache in #316.

## Changes

- The table's truelength (`HASHCOUNT`, formerly `HASHPRI`) is the exact number of bindings, maintained by insert, delete, resize, `attach()` and the load-time restore. `env.profile()` counts non-empty chains itself, so its `nchains` is now correct.
- A table holding more than 0.75 bindings per chain moves to the next size of a ladder, the first prime above 1.618 times each power of two from 2^5 to 2^30; a user-sized table moves to the first ladder size at least 1.5 times its own. The global cache uses the same check.
- `R_HashResize()` uses the hash cached in each name's CHARSXP instead of rehashing the string.
- Documentation in R Internals and `?environment`; a regression test checks the count against the contents after inserts, removals and a serialization round trip.

## Why these sizes

The first attempt doubled to 2n + 1, which from 29 gives 15 * 2^k - 1, and lookups got *slower*: a million `key<N>` names occupied 23% of the chains where a uniform hash gives 40%. `R_Newhashpjw()` is a base-16 polynomial of the characters (`h = (h << 4) + c` with the top nibble folded back), and the index is the hash modulo the size, so a modulus near a small multiple of a power of two folds the lattice of digit values onto itself. Primes do not help by themselves: 2097169, the first prime above 2^21, is just as bad, while 2000001 and 1000003 are fine. The hash cannot change, because hashed environments are serialized with their chains in place and must stay readable, so only the sizes can. Simulated occupancy relative to a uniform hash, a million names at load 0.5, six name shapes:

| sizes | `key<N>` | `v<N>` random order | 8 random letters | `pkg.fun_<N>` | `<N>` | `my_var_<N>_x` |
|---|---|---|---|---|---|---|
| 2n + 1 from 29, at 2^21 | 0.59 | 0.58 | 0.97 | 0.68 | 0.53 | 0.78 |
| prime above 1.618 * 2^k | 0.97 to 1.08 | 0.99 to 1.06 | 1.00 | 0.95 to 1.08 | 0.96 to 1.09 | 1.00 to 1.04 |

The old growth by 1.2 produced irregular sizes and avoided the problem by accident (a grown table of 864,241 slots was at 0.88 of uniform).

## Measurements

aarch64 macOS, `-O2`, `R_SYMBOL_TABLE_SIZE=2000000` so that the symbol table does not dominate (with the default symbol table the million-key figures are all about 1.7 us higher on both sides). Nanoseconds per binding:

| 1e6 bindings | before | after |
|---|---|---|
| `list2env()` into a default-size environment (grows on the way) | 890 | 325 |
| `list2env()` into a presized environment | 346 | 294 |
| `assign()` one at a time (dominated by `assign()` itself) | 2565 | 2487 |
| `mget()`, grown table | 620 | 590 |
| bindings visited per hit, grown table | 1.75 | 1.22 |
| table slots | 864,241 | 1,696,649 |

| 1e5 bindings | before | after |
|---|---|---|
| `list2env()` into a default-size environment | 410 | 160 |
| `mget()` | 320 | 250 |

`attach()` of a 200,000-element list takes 652 ms instead of 614, since the presized table now resizes once. Loading and unloading `stats4` twenty times, and R startup, are unchanged. The table takes about 16 bytes per binding instead of 4 to 8, against the 56 bytes of the binding's cons cell.

### DIFF
--- a/doc/NEWS.Rd
+++ b/doc/NEWS.Rd
@@ -109,7 +109,9 @@
       about twice as fast.  The load
       factor and the initial size of the table can be set at startup
       with environment variables \env{R_CHARSXP_CACHE_LOAD_FACTOR} and
-      \env{R_CHARSXP_CACHE_INITIAL_SIZE}, see \code{?EnvVar}.
+      \env{R_CHARSXP_CACHE_INITIAL_SIZE}, and the size of the (fixed
+      size) symbol table with \env{R_SYMBOL_TABLE_SIZE}, see
+      \code{?EnvVar}.
     }
   }
 
--- a/doc/NEWS.Rd
+++ b/doc/NEWS.Rd
@@ -106,7 +106,11 @@
       bucket), and \code{unserialize()} and hence \code{readRDS()} read
       character vectors in batches which prefetch the cache entries
       they will need.  Reading a file of a million distinct strings is
-      about twice as fast.  The load
+      about twice as fast.  The garbage collector no longer sweeps the
+      whole cache on every collection, only the buckets holding strings
+      young enough to have died, which removes a per-collection cost
+      that grew with the number of strings a session had ever created.
+      The load
       factor and the initial size of the table can be set at startup
       with environment variables \env{R_CHARSXP_CACHE_LOAD_FACTOR} and
       \env{R_CHARSXP_CACHE_INITIAL_SIZE}, and the size of the (fixed
--- a/doc/NEWS.Rd
+++ b/doc/NEWS.Rd
@@ -99,6 +99,17 @@
       \item \code{rfree1way} also accepts \code{offset} terms equal to
       the number of simulated observations, allowing to sample from more
       complex models.
+
+      \item Creating character strings is faster: the internal cache of
+      \code{CHARSXP}s now doubles its hash table at 0.75 strings per
+      bucket rather than at 85\% bucket occupancy (about 1.9 per
+      bucket), and \code{unserialize()} and hence \code{readRDS()} read
+      character vectors in batches which prefetch the cache entries
+      they will need.  Reading a file of a million distinct strings is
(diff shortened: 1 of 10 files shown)


kevinushey:
Split into a series of three standalone patches based directly on `main` (no dependency on #316): #324 (count bindings exactly, fix `env.profile()`), #325 (resize via the cached CHARSXP hashes), #326 (grow at 0.75 to prime ladder sizes). #326 supersedes this PR.

PR #324, opened 2026-09-03
count hashed environment bindings exactly

kevinushey:
# Summary

Hashed environments counted their bindings inconsistently, so `env.profile()` reported a wrong number of chains and the resize threshold compared against a mixture of two different counts. The count is now the exact number of bindings, and `env.profile()` counts non-empty chains itself.

First of a series of three split out of #320 so each piece can be presented as a standalone patch; this one is independent of the others. The series is based directly on `main` and, unlike #320, does not depend on #316.

## What was wrong

- `R_HashSet()` increments the count for every new binding: its search loop always ends at an empty chain, so the `if (ISNULL(chain))` guard never failed. A resize, `attach()` and `R_RestoreHashCount()` (after loading a saved environment) counted occupied chains instead.
- So `env.profile()$nchains` was wrong after any insert, and the resize threshold -- grow by 20% at 85% of that mixed count -- was neither an occupancy nor a load factor: measured tables sat at 1.0 to 1.2 bindings per chain, with 1.4 to 1.75 bindings visited per hit.

## Changes

- The table's truelength (`HASHCOUNT`, formerly `HASHPRI`) is the exact number of bindings, maintained by insert, delete, resize, `attach()` and the load-time restore. The growth policy itself is untouched: grow by 20% at more than 0.85 bindings per chain, which is what the insert path already measured between resizes.
- `env.profile()` counts non-empty chains itself, so its `nchains` component is now correct.
- The CHARSXP cache at the end of `envir.c` stores its count of occupied chains in the same truelength slot; it keeps the old `HASHPRI` name and its behavior is unchanged.
- Documentation in R Internals (both places describing `truelength` as the number of primary slots in use) and `?environment`; a regression test checks the count against the contents after inserts, removals and a serialization round trip.

### DIFF
--- a/doc/NEWS.Rd
+++ b/doc/NEWS.Rd
@@ -329,6 +329,13 @@
       zeros now, similarly to \code{lm()} via \code{glm.control()}; correct
       AIC and logLik() for this case, fixing \PR{16008}, thanks to \I{Bill
 	Dunlap} and \I{Mauricio Vargas}.
+
+      \item Hashed environments keep an exact count of their bindings.
+      Previously an insert counted bindings while a resize,
+      \code{attach()} and loading a saved environment counted occupied
+      chains, so \code{env.profile()} reported a wrong number of chains
+      and the resize threshold compared against a mixture of the two.
+      \code{env.profile()} now reports the number of non-empty chains.
     }
   }
 }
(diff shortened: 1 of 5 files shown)


kevinushey:
Filed upstream: https://bugs.r-project.org/show_bug.cgi?id=19158

PR #325, opened 2026-09-03
resize environment hash tables without rehashing the strings

kevinushey:
# Summary

`R_HashResize()` and `R_HashFrame()` rehashed every name's characters with `R_Newhashpjw()`, although the hash is cached in the name's CHARSXP the first time it is computed. Both now use the cached hash, through `hashIndex()`.

Second in the series split out of #320, stacked on #324 (this PR is the last commit); #326 builds on it.

## Changes

- `hashIndex()` moves out of the `USE_GLOBAL_CACHE` section: it never depended on the global cache, and `R_HashResize()` is compiled unconditionally, so a build with `USE_GLOBAL_CACHE` undefined would otherwise fail to link.
- The resize loop relinks the chain cells as before; the dead `MIKE_DEBUG` blocks and a stale comment are dropped.

### DIFF
--- a/doc/NEWS.Rd
+++ b/doc/NEWS.Rd
@@ -329,6 +329,13 @@
       zeros now, similarly to \code{lm()} via \code{glm.control()}; correct
       AIC and logLik() for this case, fixing \PR{16008}, thanks to \I{Bill
 	Dunlap} and \I{Mauricio Vargas}.
+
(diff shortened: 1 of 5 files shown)


kevinushey:
Filed upstream: https://bugs.r-project.org/show_bug.cgi?id=19159 (patch 1 of 2 there)

PR #326, opened 2026-09-03
grow environment hash tables at 0.75 bindings per chain, to prime sizes that suit the PJW hash

kevinushey:
# Summary

Hashed environments grow their table at a measured load factor, to prime sizes chosen for the PJW hash. Filling an environment with a million bindings one at a time is 2.7x faster, and lookups in large environments visit 1.2 bindings per hit instead of 1.75.

Last in the series split out of #320 (which it supersedes), stacked on #324 and #325; this PR is the last commit. Unlike #320 the series applies directly to `main`: the CHARSXP cache keeps its own growth trigger (occupied chains above 85% of the size, now written out at its one call site), so nothing here depends on #316.

## Changes

- A table holding more than 0.75 bindings per chain moves to the next size of a ladder, the first prime above 1.618 times each power of two from 2^5 to 2^30; a user-sized table moves to the first ladder size at least 1.5 times its own. Every binding on a chain that a lookup visits is usually a cache miss, so the load factor sets the cost of a lookup.
- The global symbol cache grows through the same check.
- Documentation in R Internals; the regression test from #324 now also checks the 0.75 load bound.

## Why these sizes

The first attempt doubled to 2n + 1, which from 29 gives 15 * 2^k - 1, and lookups got *slower*: a million `key<N>` names occupied 23% of the chains where a uniform hash gives 40%. `R_Newhashpjw()` is a base-16 polynomial of the characters (`h = (h << 4) + c` with the top nibble folded back), and the index is the hash modulo the size, so a modulus near a small multiple of a power of two folds the lattice of digit values onto itself. Primes do not help by themselves: 2097169, the first prime above 2^21, is just as bad, while 2000001 and 1000003 are fine. The hash cannot change, because hashed environments are serialized with their chains in place and must stay readable, so only the sizes can. Simulated occupancy relative to a uniform hash, a million names at load 0.5, six name shapes:

| sizes | `key<N>` | `v<N>` random order | 8 random letters | `pkg.fun_<N>` | `<N>` | `my_var_<N>_x` |
|---|---|---|---|---|---|---|
| 2n + 1 from 29, at 2^21 | 0.59 | 0.58 | 0.97 | 0.68 | 0.53 | 0.78 |
| prime above 1.618 * 2^k | 0.97 to 1.08 | 0.99 to 1.06 | 1.00 | 0.95 to 1.08 | 0.96 to 1.09 | 1.00 to 1.04 |

The old growth by 1.2 produced irregular sizes and avoided the problem by accident (a grown table of 864,241 slots was at 0.88 of uniform).

## Measurements

aarch64 macOS, `-O2`, `R_SYMBOL_TABLE_SIZE=2000000` (via #316's env var) so that the symbol table does not dominate (with the default symbol table the million-key figures are all about 1.7 us higher on both sides). Nanoseconds per binding:

| 1e6 bindings | before | after |
|---|---|---|
| `list2env()` into a default-size environment (grows on the way) | 890 | 325 |
| `list2env()` into a presized environment | 346 | 294 |
| `assign()` one at a time (dominated by `assign()` itself) | 2565 | 2487 |
| `mget()`, grown table | 620 | 590 |
| bindings visited per hit, grown table | 1.75 | 1.22 |
| table slots | 864,241 | 1,696,649 |

| 1e5 bindings | before | after |
|---|---|---|
| `list2env()` into a default-size environment | 410 | 160 |
| `mget()` | 320 | 250 |

`attach()` of a 200,000-element list takes 652 ms instead of 614, since the presized table now resizes once. Loading and unloading `stats4` twenty times, and R startup, are unchanged. The table takes about 16 bytes per binding instead of 4 to 8, against the 56 bytes of the binding's cons cell.

### DIFF
--- a/doc/NEWS.Rd
+++ b/doc/NEWS.Rd
@@ -329,6 +329,13 @@
       zeros now, similarly to \code{lm()} via \code{glm.control()}; correct
       AIC and logLik() for this case, fixing \PR{16008}, thanks to \I{Bill
 	Dunlap} and \I{Mauricio Vargas}.
+
+      \item Hashed environments keep an exact count of their bindings.
+      Previously an insert counted bindings while a resize,
+      \code{attach()} and loading a saved environment counted occupied
+      chains, so \code{env.profile()} reported a wrong number of chains
+      and the resize threshold compared against a mixture of the two.
+      \code{env.profile()} now reports the number of non-empty chains.
     }
   }
 }
--- a/doc/NEWS.Rd
+++ b/doc/NEWS.Rd
@@ -107,6 +107,13 @@
       \item \code{tools::add_datalist()} now returns a logical
       (invisibly, instead of \code{NULL}), indicating if it did add
       a \file{data/datalist} file (addressing parts of \PR{18015}).
+
+      \item Hashed environments grow their hash table to a prime of
+      about twice the size once it holds more than 0.75 bindings per
+      chain, rather than by 20\% at 85\% full.  Filling an environment
+      with a million bindings one at a time moves each binding about
+      1.5 times on the way instead of 5.7, and lookups in large
+      environments visit fewer bindings.
     }
   }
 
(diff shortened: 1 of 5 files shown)


kevinushey:
Filed upstream: https://bugs.r-project.org/show_bug.cgi?id=19159 (patch 2 of 2 there)

PR #328, opened 2026-09-06
free radix sort's scratch buffers when an error unwinds

kevinushey:
`do_radixsort` in `src/main/radixsort.c` allocates the per-column scratch
buffers `xsub` and `newo` with `malloc`, and frees them, along with the
file-static working buffers (`radix_xsub`, `xtmp`, `otmp`, `csort_otmp`,
`cradix_*`, the group-size stacks, `ustr`), only on its normal return. An
error signalled mid-sort unwinds past those frees. The `Error()` macro
fires when a key column past the first is of an unsupported type, and there
are allocation-failure and stack-overflow paths as well.

Because the buffers are file statics, the next call overwrites the pointers
with fresh allocations, so the leak is unbounded across calls rather than a
one-off.

## Reproducer

```r
n <- 1e6
for (i in 1:100)
    try(order(rep(1L, n), complex(n), method = "radix"), silent = TRUE)
# each call: "Arg 2 is type 'complex', not yet supported"
```

`order()` sorts by the first key, then the type of the second key is
checked inside the column loop, after `xsub`/`newo` were allocated for the
first key's largest tie group. A `complex` or `raw` second key reaches the
`Error()`. On unpatched trunk r90492 (aarch64-apple-darwin25.6.0), macOS
`leaks` reports ~11.8 MB leaked per call (12 bytes per element of the
largest tie group); patched, zero.

## A latent correctness bug falls out of the same defect

The string-grouping path marks the cached CHARSXPs via their truelength
and clears the marks only on the normal return. When an error unwinds
first, the marks are left in place, so the *next* radix-sort call aborts:

```r
try(order(c("b", "a", "b"), complex(3), method = "radix"), silent = TRUE)
sort(c("b", "a", "b"), method = "radix")
#> Error: Internal error. ustr isn't empty when starting cgroup: ustr_n=2 ...
```

## Fix

Move the whole teardown into a `radix_cleanup()` invoked from a
`CTXT_CCODE` cleanup context, so it runs on the error unwind as well as the
normal return (the idiom `do_dput` / `deparse1WithCutoff` use). The cleanup
restores the truelengths and resets `ustr_n` / `maxlen` in the same order
the normal path did; `do_radixsort`'s own local `xsub` is reached through
the context data pointer. The file needs `R_USE_SIGNALS` for the context
API, as `scan.c` and `deparse.c` already set.

Verified on a patched build: the leak is gone, `reg-tests-1c/1d/1e` pass,
the post-error grouping results match a clean process exactly and are
identical across repeated calls, and the "Internal error" abort above no
longer occurs.

Found by a static scan for resources released only on the normal return
path (a checker written for this sweep), reproduced with the macOS leaks
tool.

### DIFF
--- a/src/main/radixsort.c
+++ b/src/main/radixsort.c
@@ -24,6 +24,7 @@
 #include <config.h>
 #endif
 
+#define R_USE_SIGNALS 1 /* for the cleanup context */
 #include <Defn.h>
 #include <Internal.h>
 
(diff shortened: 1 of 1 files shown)


MichaelChirico:
does this reproduce on data.table as well?

kevinushey:
Checked data.table (dev 1.18.99 source, 1.18.2.1 installed) -- no, it doesn't reproduce, for two reasons.

`forder` no longer stamps CHARSXP truelengths (it groups strings via a separate `marks` hash table), so the truelength-corruption half can't occur; forcing an errored sort and then re-sorting strings gives correct results with no "ustr isn't empty" abort.

And the buffer-leak half is guarded by convention: `forder.c` forbids bare `error()` and routes every error through `STOP()` / `internal_error_with_cleanup()`, both of which call `cleanup()` to free the scratch first -- the discipline base R was missing here. It also front-loads type validation (complex sorts fine there), so R's "unsupported column type discovered mid-loop" trigger is an unreachable internal-error path in data.table rather than an ordinary one.

MichaelChirico:
excellent, thanks for checking! I guess I should make a mental note that fixes we make should be offered as patches here too.

PR #338, opened 2026-09-12
old.packages(): only use Built field when checkBuilt = TRUE

jeroen:
R 4.6 changed behavior such that `update.packages()` will reinstall packages if the timestamp for the `Built` field has changed.

This has surprised some users, see for example [r-universe discussion](https://github.com/orgs/r-universe-org/discussions/729) and the r-package-devel thread [recent changes to r-devel with package date checking](https://stat.ethz.ch/pipermail/r-devel/2026-March/date.html)

This PR makes the behavior opt-in based on the existing `checkBuilt = TRUE` parameter.


s-u:
@jeroen This is intentional. The problem is that checkBuilt means something entirely different that the name suggests - it it a check of the major R version which is why it defaults to FALSE. New binaries must be treated as new since, well, they are new requiring update regardless of the checkBuilt flag. If you have a repository that does binary-identical re-builds (like Debian) then you need to make sure the Built field is preserved, in all other cases a re-build is always a new version.

jeroen:
Right, the idea is to give users control on if they want to reinstall new binaries for the same package version. So rather than introducing a new parameter, I thought it would make sense to make `checkBuilt` actually do what the name suggests, i.e. enable/disable package updates based on 'built' properties (target R version, timestamp) rather than only version number. I think it makes sense to interpret it this way. Actually users were already [expecting this](https://github.com/orgs/r-universe-org/discussions/729#discussioncomment-18414527) in some sense.

eddelbuettel:
@s-u: Debian does not replace existing binary with rebuilds under the same 'version' id, ever.  

The 'reproducible builds' initiative (which is by now also multi-distro though it started with Debian) is checking for binary-identical build _ability_ but that is always done with a new test build.  Not in the distro, not replacing in-distro builds in the repository from which users download.  It is an entirely auxiliary test (similar to, say, reverse dependency checks).

For builds in the distro, Debian always suffixes the version of a new builds, and you can see e.g. '+1' and many other schemes for different reasons, see below in 'details'.   

The CRAN use is a little different as the Windows situation is architecturally different.  But I do feel that replacing what everybody else considers to be immetutable artifacts with something else and not changing the identifier (i.e. the file name) is a risky game to play.  But hey, your call.

<details>

Quoting for convenience from a Gemini summary for [this search](https://www.google.com/search?q=debian+reference+version+number+suffixes+reference&sca_esv=92a80f919d9b6710&biw=1896&bih=936&sxsrf=APpeQnu7az__jtdtcyC59FTt8-bLCeBSZw%3A1789244844781&ei=rLWlavqVL7n_ptQPkt2N2Ac&ved=2ahUKEwj6m4678OmWAxW5v4kEHZJuA3sQ4dUDegQIBhAM&uact=5&oq=debian+reference+version+number+suffixes+reference&gs_lp=Egxnd3Mtd2l6LXNlcnAiMmRlYmlhbiByZWZlcmVuY2UgdmVyc2lvbiBudW1iZXIgc3VmZml4ZXMgcmVmZXJlbmNlMgUQIRigATIFECEYoAEyBRAhGKABMgUQIRigATIFECEYoAFI5jJQrwdYpjFwA3gBkAEAmAFloAGyDKoBBDIwLjG4AQPIAQD4AQGYAhigAucMwgIKEAAYRxjWBBiwA8ICDhAAGOQCGNYEGLAD2AEBwgIXEC4Y3AYYuAYY2gYY2AIYyAMYsAPYAQHCAgUQIRifBcICBRAhGKsCwgIHECEYChigAZgDAIgGAZAGDroGBggBEAEYCZIHBDIyLjKgB4p1sgcEMTkuMrgH4QzCBwUxMC4xNMgHGYAIAQ&sclient=gws-wiz-serp)

---

Debian package version numbers follow the standard format: [epoch:]upstream_version[-debian_revision]. [1] (https://serverfault.com/questions/604541/debian-packages-version-convention) 

You can read the full specification in the Debian Policy Manual or look up technical details in the Debian Reference Manual. [1] (https://www.debian.org/doc/manuals/debian-reference/debian-reference.en.pdf), [2] (https://www.debian.org/doc/debian-policy/ch-controlfields.html)

### Core Structure Components

- epoch: An optional integer prefix (default is 0) used to fix or reset broken upstream versioning schemes.
- upstream_version: The main version number provided by the original software author.
- debian_revision: The number after the last hyphen (-), representing Debian-specific packaging, scripts, and patches. [1] (https://unix.stackexchange.com/questions/114708/what-do-the-different-version-elements-specify-in-debian-package-name), [2] (https://serverfault.com/questions/604541/debian-packages-version-convention)

### Common Suffixes and Indicators

#### Revision and Upload Suffixes (After the Hyphen)

- -X (e.g., -1, -2): Standard Debian package revision. Resets to -1 when the upstream version increases.
- -X.Y (e.g., -2.1): Non-maintainer upload (NMU) of a non-native package by someone other than the official maintainer.
- +nmuX (e.g., +nmu3): Non-maintainer upload of a native Debian package.
- +bX (e.g., +b1): Binary-only non-maintainer upload (binNMU), meaning the binary package was rebuilt for a specific architecture without changing the source code. [1] (https://wiki.debian.org/binNMU), [2] (https://readme.phys.ethz.ch/documentation/debian_version_numbers/), [3] (https://serverfault.com/questions/604541/debian-packages-version-convention)

### Upstream and Special Suffixes

- ~ (Tilde, e.g., 1.0.0~pre1): Sorts earlier than the base version. Used for pre-releases, betas, or backports so that the final release (1.0.0) naturally upgrades over them.
- ~bpoX (e.g., ~bpo12+1): Backport suffix indicating the package was rebuilt for an older Debian release.
- +dfsg: Indicates the original source code was modified by Debian to remove non-free files according to the Debian Free Software Guidelines.
- +tX: Translation updates (TDebs) added to the package. [1] (https://unix.stackexchange.com/questions/703950/what-does-a-tilde-mean-at-the-end-of-a-dependency-requirement-in-a-deb-file), [2] (https://unix.stackexchange.com/questions/230911/what-is-the-meaning-of-the-tilde-in-some-debian-openjdk-package-version-string), [3] (https://dep-team.pages.debian.net/deps/dep4), [4] (https://readme.phys.ethz.ch/documentation/debian_version_numbers/)

If you are trying to compare or evaluate a specific version string, paste the full version number here and I can break down what each part means.

</details>

s-u:
Since `checkBuilt` means something different:

> `checkBuilt: If ‘TRUE’, a package built under an earlier major.minor`
> `            version of R (e.g., ‘3.4’) is considered to be ‘old’.`

it cannot be used this way. The naming is unfortunate, because it didn't refer to the `Built` field of the packages, but rather the repository metadata record, which is different (just `R x.y`) and was only used to upgrade package between R versions, not for regular package installation, which is why it was disabled by default.

However, the "freshness" of a packages is by design defined by its built date, and binary packages can be updated requiring re-install and is very important (e.g. when updating dependent library due to CVE), so `old.packages` does and should do this by default. A repository should only be re-building packages if it wants the user to get the update, there is no other reason as it creates inconsistencies.

So we could add a new flag like `ignoreUpdates=FALSE` or something like (or `sourceVersionOnly=FALSE`?) that to allow an override, but that should be very unusual as it will result in packages not being updated when they should be, so I'm reluctant to open that can of worms as that is precisely what we want to avoid.

Eventually, we could deprecate the use of `checkBuilt` entirely given ti is a misnomer, but likely code relies on it, and it's not that important, just confusing.

jeroen:
> binary packages can be updated requiring re-install and is very important (e.g. when updating dependent library due to CVE), so old.packages does and should do this by default. A repository should only be re-building packages if it wants the user to get the update, there is no other reason as it creates inconsistencies.

I guess this is where the change of viewpoint arises, as this contradicts with what CRAN does on Windows (and technically Linux as well but those are not distributed).

In R-universe, and Bioconductor, and CRAN on Windows, packages are rebuilt on a continuous basis to refresh the check results and because R does not have a system in place to detect ABI changes in dependencies that require a rebuild. However it is unnecessary for users to keep updating these binaries, and before R version 4.6, this did not happen and I don't remember any problems. 

The case of a package needing updates due to being affected by a CVE in a dependency is in R very rare, I doubt it ever really happened (usually a CVE in e.g. openssl does not actually affect R bindings). I don't think it is a compelling argument; if it really would happen some day that an R package exposes a risk to users due to a security problem in a library, the R package should probably just bump the R package version number to force the update, instead of relying on built timestamps.

So therefore I would argue that forcing update of existing packages should be opt-in via `checkBuilt`. However if you do still feel the behavior is intended, this PR can be closed I guess. But it does make things awkward as your implementation contradicts the CRAN practice on all platforms (except for macos ;) where binaries are continuously refreshed, causing all these unnecessary updates to be pushed to users.

s-u:
> In R-universe, and Bioconductor, and CRAN on Windows, packages are rebuilt on a continuous basis to refresh the check results and because R does not have a system in place to detect ABI changes in dependencies that require a rebuild. 

That is not quite true - that was the other change in 4.6 specifically to address that issue: packages can flag that they trigger an ABI change requiring re-build of reverse dependencies which bumps the `Published` field in the source repository metadata for those packages, thus triggering a re-build (BTW: that's why it is checked in `old.packages` as well since ABI can change at R level as well so even if no native code is involved). Otherwise there is no reason to re-build a package. I think Uwe has his own way to do the re-builds and he didn't have time to use that mechanism which is why Windows binaries are rebuilt more often than they should, but that will settle eventually.

s-u:
> The case of a package needing updates due to being affected by a CVE in a dependency is in R very rare, I doubt it ever really happened (usually a CVE in e.g. openssl does not actually affect R bindings).

It does since it is statically linked so it very much requires a re-build. Also we had this case already more than once, last I remember was `libpq`. 

But the real motivation was the common breakage due to dependenency ABI changes alluded to above - recall the Matrix and TBB issues, all rev deps have to be updated together, so the only way to do that is for the binaries to be re-built and picked up.

> I don't think it is a compelling argument; if it really would happen some day that an R package exposes a risk to users due to a security problem in a library, the R package should probably just bump the R package version number to force the update, instead of relying on built timestamps.

Again it is very common and it's not possible (and makes no sense whatsoever) to bump the source version. Also see above, it would require massive changes to the source repo for no good reason.

s-u:
I am open to suggestions to improve the system, but it was designed specifically to cover all the issues you listed above. If it need more documentation/tools or similar, that's fine, we can talk about that, but I do think it fundamentally works and solved the main issue we had before where you could get broken library with `update.packages(type="binary")` because rev deps were not updated where they needed to be.

jeroen:
> That is not quite true - that was the other change in 4.6 specifically to address that issue: packages can flag that they trigger an ABI change requiring re-build of reverse dependencies which bumps the `Published` field in the source repository metadata for those packages, thus triggering a re-build (BTW: that's why it is checked in `old.packages` as well since ABI can change at R level as well so even if no native code is involved). Otherwise there is no reason to re-build a package. I think Uwe has his own way to do the re-builds and he didn't have time to use that mechanism which is why Windows binaries are rebuilt more often than they should, but that will settle eventually.

OK I have never heard of this. Suppose I want to implement the same mechanism in R-universe. How exactly do packages flag such a breaking/ABI change to their reverse dependencies, so that I know on the server side I need to trigger rebuilds of revdeps? I don't see this mechanism mentioned in WRE? Also for packages with S4 classes it would be nice to know when some changed class/method definition requires rebuilding the revdeps, but afaik there is no mechanism for this?

> Otherwise there is no reason to re-build a package. 

Well there is always the reason to refresh CMD check results, which can change due to all sorts of circumstances. Of course you can do a new `R CMD check` without publishing the new binaries, but it might make sense to keep them in sync.

PR #339, opened 2026-09-12
bound recoding of serialized strings

kevinushey:
A version-3 serialization stream can label a length-delimited string as the
platform's `wchar_t` encoding. When conversion failed, the warning path passed
that buffer to `reEnc3()`, which inferred a wide string from the encoding name
and called `wcslen()` beyond the serialized string's bounds.

Add `reEnc4()` for callers with an explicit byte length and use it for
length-delimited serialized strings, `CHARSXP` diagnostics, and buffered wide
string conversions. Keep `reEnc3()` for NUL-terminated inputs whose lengths
are not otherwise available. Also reject embedded NULs in version-3 encoding
names in both `unserialize()` and `infoRDS()`.

Validation:

- clean out-of-tree R build
- `reg-tests-1b.R`
- `reg-tests-1b.R` with an AddressSanitizer-built core
- original ClusterFuzz testcase and a direct platform-wide-encoding trigger
  under AddressSanitizer

### DIFF
--- a/src/gnuwin32/Rdll.hide
+++ b/src/gnuwin32/Rdll.hide
@@ -544,6 +544,7 @@
  Rf_promiseArgs
  Rf_qchisq_appr
  Rf_reEnc3
+ Rf_reEnc4
  Rf_readS3VarsFromFrame
  Rf_regcomp
  Rf_regerror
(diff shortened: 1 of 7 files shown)


kevinushey:
Filed upstream: https://bugs.r-project.org/show_bug.cgi?id=19047 (comment + patch on the existing malicious-.rds bug).

PR #340, opened 2026-09-15
validate serialized ALTREP class metadata

kevinushey:
A malformed ALTREP stream can supply arbitrary class metadata. In particular,
an invalid serialized base type such as `8357` reaches `type2char()`, indexes
beyond `Type2Table`, and crashes while formatting the type-mismatch warning.

Validate the class metadata structure, symbol fields, scalar integer type
field, and supported ALTREP base type before class lookup. Malformed inputs now
raise a regular R error. Regression coverage includes the ClusterFuzz type
value, a wrong-typed field, and a zero-length field.

Validation:

- clean out-of-tree build
- `make reg-tests-1e.Rout`
- ClusterFuzz testcase `unserialize-6318914150400000`
- normal integer, real, and string ALTREP serialization round trips


kevinushey:
Filed upstream: https://bugs.r-project.org/show_bug.cgi?id=19047

